PR #2, opened 2020-04-06
Preparing for 4x3 update

pjanck:
Lots of small fixes.


hechth:
I'm getting this error in CMake.

CMake Error at C:/Qt/5.12.1/msvc2017_64/lib/cmake/Qt5Core/Qt5CoreMacros.cmake:376 (target_link_libraries):
  The keyword signature for target_link_libraries has already been used with
  the target "BlueFramework.Application".  All uses of target_link_libraries
  with a target must be either all-keyword or all-plain.

  The uses of the keyword signature are here:

   * C:/vs_projects/Open-Infra-Platform/_deps/blue_framework-src/Application/CMakeLists.txt:87 (target_link_libraries)

hechth (changes requested):
Please redo the Qt deployment and copying.

pjanck:
This problem is connected to [Issue 1](https://github.com/tumcms/Blue-Framework/issues/1) and [Issue 2](https://github.com/tumcms/Blue-Framework/issues/2) from [Blue-Framework](https://github.com/tumcms/Blue-Framework).
.

PR #17, opened 2020-04-10
Created InfoDoxygenSetupOIP.pdf

ghost:
Doxygen setup for Open Infra Platform info file added to documentation.


pjanck (approved):
It would be better to have this as a markdown document, checked-in and editable. Currently the source file of the PDF is not part of the repository. But for a first, this is great.

PR #21, opened 2020-04-13
Rewrote the readStepData function of ExpressContainer completely

pjanck:
It now supports lists of lists and takes care of (( ... ),( ... ), ... ) as expected.


hechth:
Possible improvement: break down method into several sub functions.

PR #22, opened 2020-04-13
Refactoring of ifc converter

pjanck:
Started refactoring IfcConverter. Nowhere near finish, but there's a start.


hechth (approved):
* Introduced ConverterBaseT class which now handles unit conversion and geometry settings
* Improved and unified logging output using `getErrorLog` function and `BLUE_LOG(trace)`
* Implemented IfcTriangulatedFaceSet geometry representation¨
* Refactored all converters to fit new unit conversion system

Issue #26, opened 2020-04-14
Doxymentation directory


pjanck:
Currently, doxymentation is being built to "Documentation\doxymentation\html" within the repository. Change this to be within the "build" folder as chosen in CMake GUI.

pjanck:
See #65 .

PR #65, opened 2020-04-22
Doxymentation directory. Resolves issue #26.

ghost:
Changed location of build directory to ${CMAKE_CURRENT_BINARY_DIR}/Doxymentation.

**New issue:** 
In the process noticed that CMake doesn't want to find the dot and doxygen executables when configuring CMake for Doxygen the first time (first time selecting DOXYGEN_GENERATE_DOCUMENTATION) - the paths have to be added manually. Something in the order of find_path(...) and find_package(...) doesn't allow it to be found yet - has to be solved.


pjanck:
Fixes #26 .

PR #91, opened 2020-04-25
Doxymentation of generated ifc files

pjanck:
Generated files now have doxymentation filled in (entities, that is).


pjanck:
Also fixes #92 .

hechth (approved):

PR #99, opened 2020-04-27
Corrected dealing with IfcPolyline.

### LINKED ISSUE #97: Error in IFC geometry
After updating the code to commit 4a93ac286dd41566cfaa0504f89fc09e0daf1844, I get errors in the geometry when loading an ifc file.

What I get
image

What I expect
image

pjanck:
It now looks like this:
grafik


kyq:
Working now! Fixes #97  
image

PR #101, opened 2020-04-29
Colour ifc objects

kyq:
Assign (random) colours to each object (i.e. each mesh in meshSet) in an ifc model.  Applied to wall objects


pjanck:
This code is here for reference. This will be included in a better way when the coloring of all objects will be addressed.

pjanck:
Archive. Saved for further reference.

PR #102, opened 2020-04-29
Dealing with 98

### LINKED ISSUE #98: Zoom to the content of IFC file not working
When loading point cloud data, the view moves to the loaded content immediately. This used to work on IFC data as well, but is not anymore.

pjanck:
Fixes #98 . 

The IFC model now provides a bounding box, so that the viewport can react appropriately. The geometry is also moved to (0,0,0), similar to the point cloud handling.

Bounding box should be a -must-have- attribute of any model, if we are to extend to multimodel support.


kyq:
Working well. IFC models are properly positioned and shown.

Fixes #98.

Core/src/IfcGeometryConverter/ConverterBuw.h
```diff
@@ -43,22 +43,63 @@ namespace OpenInfraPlatform
...
+				std::string toString() const {
+					return "min: (" + std::to_string(min_.x()) + ", " + std::to_string(min_.y()) + ", " + std::to_string(min_.z())
+					   + ") max: (" + std::to_string(max_.x()) + ", " + std::to_string(max_.y()) + ", " + std::to_string(max_.z()) + ")";
+				}
+				buw::Vector3d center() const { return 0.5 * (min_ + max_); }
+				buw::Vector3d min() const { return min_; }
+				buw::Vector3d& min() { return min_; }
+				buw::Vector3d max() const { return max_; }
+				buw::Vector3d& max() { return max_; }
+			};
```
hechth: This belongs into an own file or own class - best as a extension maybe to the carve bounding box or by having an own interface which has an implementation that uses carve

so oip::IBoundingBox -> BoundingBoxImpl which uses carve bounding box and forwards functionality.

hechth (changes requested):
Please rework the bounding box to incorporate the functionality from carve or have a custom interface that has an implementation which forwards functionality.

Core/src/IfcGeometryConverter/ConverterBuw.h
pjanck: done.

kyq:
Works fine with both clouds and ifc models

PR #113, opened 2020-05-06
Improving converters

pjanck:
Introduced isOfType<..>(..) function for easier code writing and run-time checking.

### DIFF
--- a/Core/src/IfcGeometryConverter/ConverterBase.h
+++ b/Core/src/IfcGeometryConverter/ConverterBase.h
@@ -40,6 +40,8 @@
 #include "GeometrySettings.h"
 #include "UnitConverter.h"
 
+#include "EXPRESS/EXPRESSReference.h"
+
 namespace OpenInfraPlatform
 {
 	namespace Core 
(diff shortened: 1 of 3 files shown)


Core/src/IfcGeometryConverter/ConverterBase.h
```diff
@@ -59,26 +59,36 @@ namespace OpenInfraPlatform
...
 
+				/**
+				 * \brief Tests if a pointer to \c Is can be casted to a pointer to \c Test.
+				 *
+				 * \param[in] object a \c std::shared_ptr of type \c Is
+				 * 
+				 * \returns \c true if dynamic_cast returns anything else but \c nullptr\n \c false otherwise
+				 */
+				template <class Test, class Is>
+				bool isOfType(const std::shared_ptr<Is>& object) { return std::dynamic_pointer_cast<Test>(object) != nullptr; }
```
hechth, line 90: This can maybe even be improved so that not the full class name has to be provided as `typename IfcEntitiyTypesT::IfcLinearPlacement` but simply `IfcLinearPlacement`

hechth (approved):

PR #122, opened 2020-05-13
WIP: Refactor representation converter

pjanck:
It compiles, but at runtime, gibberish comes out. 

The problem happens for example while converting a `IfcCartesianPoint` in `PlacementConverterT::convertIfcCartesianPoint(..)`.
I suspect that by passing `const EXPRESSReference<>&` , the function cannot access the non-const `operator->()` and with that the non-const `lock()` function. Through that, the `operator=()` is not called if the `weak_ptr` is expired.


pjanck:
This is now a branch on the official repository.

PR #145, opened 2020-05-21
Added attributes 1-2 to sectioned solid

mauplatteau:
now it gets the attributes directrix and cross sections


pjanck (changes requested):
Yes, that's the correct direction.

pjanck:
Archive. New pull request schould be made following our conversation today (when ready).

PR #159, opened 2020-06-12
Update PointCloud.cpp

fbosche:
Change ""bin"" to "extension" to enable loading point clouds with any supported extension.


fbosche:
Error

Issue #139, opened 2020-05-20
[REF] SplineConverter.h


hechth:
**File which is being refactored**
IfcGeometryConverter\SplineConverter.h

**Functions**

List all functions in the file which have to be refactored with possible code snippets!

- [ ] Listed all functions in the issue.

christophKaiser:
Functions in refactoring:
- [x] convertIfcBSplineCurve including its specialized template function
- [ ] convertIfcBSplineSurface including its specialized template function
- [x] computeBSplineBasisFunctions
- [ ] computeBSplineSurface
- [x] computeBSplineCurve

[Updated 2020-07-21]

PR #178, opened 2020-07-11
Refactoring splineConverter.h

christophKaiser:
### Content of description
1. [ What's done until now ](#WhatsDone)
1.1. [ Overview ](#WhatsDoneOverview)
1.2. [ In more detail ](#WhatsDoneDetail)
2. [ Questions to discuss ](#Discuss)
3. [ Pending work ](#ToDo)

<a name="WhatsDone"></a>
### What's done until now
The github view 'Files changed' doesn't give a well overview about what happened, because some commented out code was reactivated and changed. As well, some code was moved into subfunctions. The changes are better comprehensible by the commits step by step.

<a name="WhatsDoneOverview"></a>
#### Overview
The old commented out code from the template function `convertIfcBSplineCurve` at the end of the file was moved inside the usual function inside the class. As well `EXPRESSReference` was introduced and the code was adjusted that the compiler builds it successfully.
Furthermore, parts of the functions `convertIfcBSplineCurve` and `computeBSplineCurve` were outsourced into smaller subfunctions to provide a better readability of the code. According to other converter classes, `convertIfcBSplineCurve` got a new branch-based decision structure for the four cases IfcRationalBSplineCurveWithKnots, IfcBSplineCurveWithKnots, IfcBSplineCurve (with special condition) and unknown types. This leads to a split of `computeBSplineCurve` into two compute functions, one for IfcBSplineCurveWithKnots and one for IfcRationalBSplineCurveWithKnots.
After a brief check, I can't see a required refactoring of the function `computeBSplineBasisFunctions`.

<a name="WhatsDoneDetail"></a>
#### In more detail
- Reactivation of the function `convertIfcBSplineCurve`:
With commit c115f07 code from the commented-out template function was pasted into the function `convertIfcBSplineCurve`. This function is called by outside the class.
With commit 45a2f52 this legacy code was modified until the compiler runs through `convertIfcBSplineCurve` without errors. As well, `std::shared_ptr` was changed to `EXPRESSReference`.
With commit ffc4eed a few variables were renamed according to the definitions in https://standards.buildingsmart.org/IFC/DEV/IFC4_3/RC1/HTML/schema/ifcgeometryresource/lexical/ifcbsplinecurve.htm and its ifc-subclasses. This should provide a better understanding and prevent confusion in future.
In commit 5733df1 EXPRESSReference.`isOfType` was applied to the if-conditions.
- Reactivation of the function `computeBSplineCurve`:
According to commit ffc4eed (see above), in commit bdb377a one variable name was renamed here as well.

&rarr; With this changes the old code should work to convert a B-spline, and EXPRESSReference is applied. Next changes should make it easier to read and understand the code.

- Refactoring of the function `convertIfcBSplineCurve`:
In commit 29030be the functions `loadWightData` and `loadKnotArray` were created. Some code from `convertIfcBSplineCurve` was moved into these new functions. In `convertIfcBSplineCurve` only one line of function call left. Later in commit 4c6a04b, slight changes in `loadKnotArray` were done.
In commit b49de9f the 'waterfall' like working structure of `convertIfcBSplineCurve` was replaced by a 'branch' like structure (similar to other converter classes of the project). With this new decision structure, it's easier to realise the four cases IfcRationalBSplineCurveWithKnots, IfcBSplineCurveWithKnots, IfcBSplineCurve (with special condition) and unknown types.
- Refactoring of the function `computeBSplineCurve`:
In commit 65fc6b6 the function `obtainKnotRange` was created. Some code from `computeBSplineCurve` was moved into this new function. In `computeBSplineCurve` only a few lines of variable declarations and the function call left.
In commit e35da99 the function type of `computeBSplineBasisFunctions` was changed from `void` to `std::vector<double>`. At the function call in `computeBSplineCurve` it's clearer what the return variable is. In commit 1dd3493 the same happened to `computeBSplineCurve`.
In commit 389ad1d the function `obtainProperties` was introduced which holds [temporary] fix properties of the curve calculation at one place. This should prevent declaration of the same information at multiple places, which will come with one of the next commits.
- Refactoring of the function `computeBSplineCurve` - the split in two functions, one for B-splines with Knots, one for rational B-splines:
Commit 7541204 adds the functions `computeIfcBSplineCurveWithKnots` and `computeIfcRationalBSplineCurveWithKnots`, their code was copied 1:1 from `computeBSplineCurve`. This is an own commit to make further changes inside the new functions clearer.
In commit b2cae32 all parts of code which are for rational B-splines were deleted in `computeIfcBSplineCurveWithKnots`.
In same way, commit 260b35f removed all parts of code which are for [normal] B-splines in `computeIfcRationalBSplineCurveWithKnots`.
In commit 87377b5 a few 'one time used' variables were replaced through using the original variable instead.

[Update 2020-07-16]
The suggested function call GeomSettings()->getPrecision() was added in commit 7096876.
The inheritance of ConverterBaseT was introduced in the constructor with commit c499f83. According to this, the function call in CurveConverterT was modified in commit c09bd83.
[Update end]

[Update 2020-07-19]
Parts of Update 2020-07-16 were reverted according to pjanck's change requests.
Doxygen comments were added in commit f048299.
Functions were moved and deleted according to the announcement in 'Pending work'. Small fixes were done. In [this](https://github.com/tumcms/Open-Infra-Platform/pull/178#issuecomment-660629477) comment the update is described in more detail.
[Update end]

<a name="Discuss"></a>
### Questions to discuss
- [answered [here](https://github.com/tumcms/Open-Infra-Platform/pull/178#issuecomment-657186427)] If an IfcBSplineCurve has no ifc-subtypes, and if the number of control points is equal to the order ( = degree + 1 ), there are enough information to calculate a Bézier Curve. With a special pre-set of the knot array, the function `computeIfcBSplineCurveWithKnots` can be used. It should be discussed whether to implement that because it isn't mentioned in the ifc documentation. 
- [answered [here](https://github.com/tumcms/Open-Infra-Platform/pull/178#commitcomment-40532962)] In commit 4c6a04b a compatibility check of two data of the ifc entity was adopted from the legacy code. If the check comes to an invalid compatibility, an exception `throw oip::UnhandledException("Function convertIfcBSplineCurve::loadKnotArray: Knot multiplicity does not correspond number of  distinct knots; unable to construct a knot array.")` will be triggered. At this point, it should be discussed, whether the check is necessary and reasonable - or in other words, whether wrong data can be inside the ifc entity. Furthermore, I'm not sure, whether that's the right kind of exception.
- [answered [here](https://github.com/tumcms/Open-Infra-Platform/pull/178#issuecomment-657186427)] In commit 29030be `std::transform` is used. I'm not sure, whether `[](auto &it) { return it; }` converts the ifc value type to an usual C++ data type. At least, the compiler has no issue in the way how it's used at the moment.

<a name="ToDo"></a>
#### Now:
- [x] Add doxygen commands to `convertIfcBSplineCurve` and all functions which are called from there.
- [ ] Test with test data, whether the converters of B-splines and rational B-splines work fine after refactoring.
- [x] Delete the function `computeBSplineCurve`, because it isn’t used anymore.
- [x] Move the function `computeBSplineBasisFunctions` into private.

#### In far future:
- [ ] Refactoring of `convertIfcBSplineSurface` and `computeBSplineSurface`.


pjanck:
Beautiful! I commend you for the detailed description and analysis of your work.

Notes and answers:
- I've left you a few tips on how to improve your code for some lines. If you'll have questions, please use the GitHub functionality of threads.
- ÌfcBSplineCurve` is an abstract entity - see [documentation](https://standards.buildingsmart.org/IFC/DEV/IFC4_3/RC1/HTML/link/ifcbsplinecurve.htm). Therefore, an instance of it cannot be saved within an IFC file. No need to worry about that.
- The check is technically not necessary, but I would leave it there. Better to throw one exception too many than to not know why a program crashes. The correct type of exception is noted in my comment to the code.
- about `std::transform` - the way it is currently coded is not wrong. However, with the change to the Express generation, one can simply use the arrays present in the IFC entity. Suggestion: Leave it as it is for now.
- I suggest that you do not refactor `...surface` functions (yet). Focus on your thesis' task. The refactor of those functions can be done later.

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -39,7 +39,13 @@ namespace OpenInfraPlatform
 			class SplineConverterT : public ConverterBaseT<IfcEntityTypesT> // TODO 2020.04.08.: spline converter does not apply length corrections
 			{
 				public:
-					SplineConverterT() {}
+					SplineConverterT(
+						std::shared_ptr<GeometrySettings> geomSettings,
+						std::shared_ptr<UnitConverter<IfcEntityTypesT>> unitConverter)
+						:
+						ConverterBaseT<IfcEntityTypesT>(geomSettings, unitConverter) 
+					{}
```
christophKaiser, line 57: The pointer to PlacementConverterT was added in Commit 9078d7d

christophKaiser:
### Update - in detail about new commits
According to pjanck's comment, in commit 024b3a1 the exception type in the function `loadKnotArray` was changed. As well, the entity which causes the exception is given to the call.

The suggested function call `GeomSettings()->getPrecision()` was added in commit 7096876. 
This caused an error (see [this](https://github.com/tumcms/Open-Infra-Platform/pull/178#commitcomment-40618456) comment), which is fixed with the commits c499f83, d1855bf, e62cf7c, d3ed525, 9078d7d and 75fe6bf. In concision: The constructor of `SplineConverterT` gets pointers to `GeometrySettings`, `UnitConverter` and `PlacementConverterT`, and calls the inherited `ConverterBaseT`; the pointer to `PlacementConverterT` is stored in a protected variable. The new `SplineConverterT` constructor is called in `RepresentationConverterT`. Some constructure calls in other converters were modified according to this change (see description of commit d1855bf for more detail).
Include statement `#include "ConverterBase.h"` was added to SplineConverter.h (commit e62cf7c)
`static` was removed from all functions in `SplineConverterT` (commit d3ed525).
With this changes the caused error was fixed.

christophKaiser:
With the update above (particularly commit d1855bf), there are changes in ProfileConverterT, which might have influence to contributor SamuilsRulovs as well?

christophKaiser:
Updated top description of the pull request.

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -519,7 +521,7 @@ namespace OpenInfraPlatform {
 						// convert the control points
 						std::vector<carve::geom::vector<3>> splinePoints = convertIfcCartesianPointVector(bspline_curve->ControlPointsList);
 
-						SplineConverterT<typename IfcEntityTypesT>::convertIfcBSplineCurve(bspline_curve, splinePoints, targetVec);
+						splineConverter->convertIfcBSplineCurve(bspline_curve, splinePoints, targetVec);
 						return;
 
 						// TO DO: Implement IfcBSplineCurveWithKnots
```
pjanck: Please delete these TODOs, if this is done with the convertBSplineCurve function.

Core/src/IfcGeometryConverter/SplineConverter.h
pjanck: If the placementConverter is not needed anywhere in the SplineConverter, then there is no need to have it.

Core/src/IfcGeometryConverter/RepresentationConverter.h
```diff
@@ -784,6 +786,7 @@ namespace OpenInfraPlatform {
 
 				// std::shared_ptr<StylesConverter>	stylesConverter;
 				std::shared_ptr<PlacementConverterT<IfcEntityTypesT>> placementConverter;
+				std::shared_ptr<SplineConverterT<IfcEntityTypesT>> splineConverter;
```
pjanck: Since it is not needed anywhere in the class, there is no need for a member variable.

Core/src/IfcGeometryConverter/ProfileConverter.h
```diff
@@ -1652,6 +1655,7 @@ namespace OpenInfraPlatform {
 		protected:
 
 			std::shared_ptr<PlacementConverterT<IfcEntityTypesT>> placementConverter;
+			std::shared_ptr<SplineConverterT<IfcEntityTypesT>> splineConverter;
```
pjanck: If CurveConverter gets splineconverter removed as member, these can be removed as well.

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -519,7 +521,7 @@ namespace OpenInfraPlatform {
 						// convert the control points
 						std::vector<carve::geom::vector<3>> splinePoints = convertIfcCartesianPointVector(bspline_curve->ControlPointsList);
 
-						SplineConverterT<typename IfcEntityTypesT>::convertIfcBSplineCurve(bspline_curve, splinePoints, targetVec);
+						splineConverter->convertIfcBSplineCurve(bspline_curve, splinePoints, targetVec);
```
pjanck:
```suggestion
						SplineConverterT splineConverterT(GeomSettings(), UnitConvert());
						splineConverter->convertIfcBSplineCurve(bspline_curve, splinePoints, targetVec);
```
Then there is no need for a member variable.

Core/src/IfcGeometryConverter/ProfileCache.h
```diff
@@ -95,6 +97,7 @@ namespace OpenInfraPlatform {
 				protected:
 
 					std::shared_ptr<PlacementConverterT<IfcEntityTypesT>> placementConverter;
+					std::shared_ptr<SplineConverterT<IfcEntityTypesT>> splineConverter;
```
pjanck: Similarly, remove the member variable, as it is not needed.

pjanck (changes requested):
Good idea, but in the end, a better solution would be to simply instantiate a SplineConverterT in CurveConverter where needed. Sorry to not have caught it sooner ...

christophKaiser:
"simply instantiate a SplineConverterT in CurveConverter where needed" (pjanck)
According to this, in far future (when refactoring goes to this point) FaceConverter will initialize in same way a local instance of SplineConverterT as well to convert a splineSurface.

Core/src/IfcGeometryConverter/CurveConverter.h
christophKaiser: Done in commit ba0ea5d

Core/src/IfcGeometryConverter/CurveConverter.h
christophKaiser: Done in commit c09bd83

Core/src/IfcGeometryConverter/RepresentationConverter.h
christophKaiser: Done in commit b13e888

Core/src/IfcGeometryConverter/ProfileConverter.h
christophKaiser: Done in commit b13e888

Core/src/IfcGeometryConverter/ProfileCache.h
christophKaiser: Done in commit b13e888

christophKaiser:
### Update
pjanck's change requestes are done in commits ba0ea5d, 02cbcd8, 603494a, 26c6e70, b13e888 and c09bd83.
Commit f048299 adds doxygen comments to the SplineConverter - at this point, I will be happe if you take a look at my [small] english skills as well.
Commits a40f180 and eeab7b7 made structural changes (move to private and delete unused stuff) which were announced in 'Pending work'.
Commits 0fa0924 and f21a2f0 added throw exceptions to the code.

christophKaiser:
Updated top description of the pull request.
In 'Pending work - now' only the test with test data left. According to this, I changed the pull request from 'draft' to 'ready for review' with the intention to be merged after successful review.

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -43,19 +47,27 @@ namespace OpenInfraPlatform
...
+						std::shared_ptr<UnitConverter<IfcEntityTypesT>> unitConverter)
 						:
-						ConverterBaseT<IfcEntityTypesT>(geomSettings, unitConverter),
-						placementConverter(pc)
+						ConverterBaseT<IfcEntityTypesT>(geomSettings, unitConverter) 
 					{}
 
 					virtual ~SplineConverterT() {}
 
+					/*! \brief Converts \c IfcBSplineCurve subtypes to a vector of curve points, which can be print on the screen.
```
pjanck:
```suggestion
					/*! \brief Converts \c IfcBSplineCurve subtypes to an array of curve points, which can be rendered in a viewport.

```

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -43,19 +47,27 @@ namespace OpenInfraPlatform
...
+					 *
+					 * \param[in]	splineCurve		\c IfcBSplineCurve entity to be converted.
+					 * \param[in]	controlPoints	A vector of the B-Spline control points, must be obtain from the \c IfcBSplineCurve entity.
+					 * \param[out]	loops			The vector of curve points, which can be print on the screen.
+					 */
 					void convertIfcBSplineCurve(
 						const EXPRESSReference<typename IfcEntityTypesT::IfcBSplineCurve>& splineCurve,
 						const std::vector<carve::geom::vector<3>>& controlPoints,
-						std::vector<carve::geom::vector<3>>& loops)
+						std::vector<carve::geom::vector<3>>& loops) throw(...)
```
pjanck:
```suggestion
						std::vector<carve::geom::vector<3>>& loops) const throw(...)
```
The function does not change the state of the SplineConverter class' instance, therefore it can be made `const`.

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -43,19 +47,27 @@ namespace OpenInfraPlatform
...
 					virtual ~SplineConverterT() {}
 
+					/*! \brief Converts \c IfcBSplineCurve subtypes to a vector of curve points, which can be print on the screen.
+					 *
+					 * This converter can handle
+					 * \c IfcBSplineCurveWithKnots and
+					 * \c IfcRationalBSplineCurveWithKnots, which are subtypes of IfcBSplineCurve.
+					 *
+					 * \param[in]	splineCurve		\c IfcBSplineCurve entity to be converted.
+					 * \param[in]	controlPoints	A vector of the B-Spline control points, must be obtain from the \c IfcBSplineCurve entity.
```
pjanck: If the points can be obtained from the `splineCurve` parameter itself, why the need to provide it to the function? The function can just obtain them itself. I don't see the reason behind it.

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -102,236 +116,40 @@ namespace OpenInfraPlatform
...
-							}
-
-							curvePoints.push_back(point);
-
-							t += step;
-						}
-
-						return curvePoints;
+						// not implemented, code in commented out part at the end of the file
+						throw oip::UnhandledException("IfcBSplineSurface isn't implementet at the moment");
```
pjanck: Please provide the ExpressReference to the exception - similarly to how it is done in line 110.

Core/src/IfcGeometryConverter/SplineConverter.h
pjanck: This can probably be applied to all functions of the class.

christophKaiser:
One commit which fixes a few warning messages will be add soon (today).

pjanck (changes requested):
Small finese to correct, otherwise very exemplary work!

christophKaiser:
> 
> 
> Please use `size_t` for this :)
(from @pjanck)

Do you meant like in the example of [C++ reference](https://en.cppreference.com/w/cpp/types/size_t)?
`for (std::size_t i = 0; i != a.size(); ++i)`

Maybe with a small change in the condition to the following?
`for (std::size_t i = 0; i < a.size(); ++i)`

pjanck:
Yes. I think that `size_t` is used quite often in the source code. Another option is to use the *for each* iterators:

    for( auto el : a )

The `el` has the same meaning as `a[i]` in your proposals.

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: Done in commit 87c2c9b
As well, adopted to all places where it's almost the same phrase.

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: Done in commit 27331ec

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: Suggestion of change in commit 754dc35
Until now, the control points were loaded in CurvConverter. I was carefully with changes in the interaction with CurveConverter, where `convertIfcBSplineCurve` is called.

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: Done in commit 73bccc9

pjanck (approved):

pjanck:
Fixes #139 .

PR #182, opened 2020-07-15
[REF] UnitConverter.h

SamuilsRulovs:
- Replaced std::shared_ptr with EXPRESSReference
- included "EXPRESS/EXPRESSReference.h" library
- Added InconsistentModellingExceptions

### DIFF
--- a/Core/src/IfcGeometryConverter/UnitConverter.h
+++ b/Core/src/IfcGeometryConverter/UnitConverter.h
@@ -23,6 +23,7 @@
 #include <memory>
 #include <vector>
 
+#include "EXPRESS/EXPRESSReference.h"
 #include "BlueFramework/Core/Diagnostics/log.h"
 
 namespace OpenInfraPlatform 
@@ -53,7 +54,7 @@ namespace OpenInfraPlatform
 
 				\param[in] project A pointer to the IfcProject entity within the model.
 				*/
-				void setIfcProject(const std::shared_ptr<typename IfcEntityTypesT::IfcProject>& project)
+				void setIfcProject(const EXPRESSReference<typename IfcEntityTypesT::IfcProject>& project) 
 				{
 					if(!project->UnitsInContext) {
 						throw oip::InconsistentModellingException(project, "project->UnitsInContext is not defined");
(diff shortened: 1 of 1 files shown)


Core/src/IfcGeometryConverter/UnitConverter.h
```diff
@@ -72,8 +73,11 @@ namespace OpenInfraPlatform
 				*/
 				double getFactorFor(const typename IfcEntityTypesT::IfcUnitEnum& unit_type) const
 				{
-					if (!m_unit_assignment)
-						return 1.;
+					
+					if (!m_unit_assignment) {
+						throw oip::InconsistentModellingException(m_unit_assignment, "m_unit_assignmentt is not defined");
+					}
+
```
pjanck: Leave it as it was. This exception is already handled in `setIfcProject`.

Core/src/IfcGeometryConverter/UnitConverter.h
```diff
@@ -121,8 +125,9 @@ namespace OpenInfraPlatform
 				*/
 				double getFactorFor(const typename IfcEntityTypesT::IfcDerivedUnitEnum& unit_type) const
 				{
-					if (!m_unit_assignment)
-						return 1.;
+					if (!m_unit_assignment) {
+						throw oip::InconsistentModellingException(m_unit_assignment, "m_unit_assignmentt is not defined");
+					}
```
pjanck: Similar to above.

pjanck (changes requested):
There are some `.lock()` function calls in the file. Since you've changed the functions to now take `EXPRESSReference` instead of `std::shared_ptr`, these can be almost everywhere removed.

Core/src/IfcGeometryConverter/UnitConverter.h
SamuilsRulovs: Resolved: e854d00

Core/src/IfcGeometryConverter/UnitConverter.h
SamuilsRulovs: Resolved: e854d00

pjanck (approved):

PR #184, opened 2020-07-15
[REF] PlacementConverter.h (WIP)

SamuilsRulovs:
Refactored functions (Replaced shared_ptr with EXPRESSReference; Added exceptions):
-  "getWorldCoordinateSystem"

### DIFF
--- a/Core/src/IfcGeometryConverter/PlacementConverter.h
+++ b/Core/src/IfcGeometryConverter/PlacementConverter.h
@@ -467,7 +467,7 @@ namespace OpenInfraPlatform {
                         return convertIfcObjectPlacement(local_placement->PlacementRelTo.get(), alreadyApplied);
                     }
                     else {
-                        BLUE_LOG(warning) << "Context based local placement computation not supported.";
+						throw oip::InconsistentModellingException(local_placement, "Context based local placement computation not supported.");
                         //TODO Georeferencing
                         // If reference to Object is omitted, then the local placement is given to the WCS, established by the geometric representation context.
                         //carve::math::Matrix context_matrix( carve::math::Matrix::IDENT() );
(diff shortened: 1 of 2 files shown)


Core/src/IfcGeometryConverter/PlacementConverter.h
```diff
@@ -467,7 +467,7 @@ namespace OpenInfraPlatform {
                         return convertIfcObjectPlacement(local_placement->PlacementRelTo.get(), alreadyApplied);
                     }
                     else {
-                        BLUE_LOG(warning) << "Context based local placement computation not supported.";
+						throw oip::InconsistentModellingException(local_placement, "Context based local placement computation not supported.");
```
pjanck: Reverse this. Otherwise no file can be read anymore.

Core/src/IfcGeometryConverter/PlacementConverter.h
```diff
@@ -826,61 +826,61 @@ namespace OpenInfraPlatform {
...
-                        }
+					if (context.expired())
+						throw oip::ReferenceExpiredException(context);
 
-                        std::shared_ptr<typename IfcEntityTypesT::IfcGeometricRepresentationContext> geom_context =
-                            std::dynamic_pointer_cast<typename IfcEntityTypesT::IfcGeometricRepresentationContext>(context);
-                        if(!geom_context) {
+					if (!context) {
+						throw oip::InconsistentModellingException(context, "context is not defined");
+					}
```
pjanck: You can't first test, if it is expired, and afterwards, if it is there. Doesn't make sense. Just delete these lines.

pjanck (changes requested):
Basically: revert all changes. Here, not only refactoring has to happen, but major changes in handling this. I need to do it ...

Core/src/IfcGeometryConverter/RepresentationConverter.h
```diff
@@ -149,13 +149,6 @@ namespace OpenInfraPlatform {
 							BLUE_LOG(warning) << ex.what();
 							continue;
 						}
-						catch (const oip::InconsistentModellingException& ex)
-						{
-							// write the error to the console
-							BLUE_LOG(warning) << representation->getErrorLog() + ": Nothing is shown - sth wrong with IFC model.";
-							BLUE_LOG(warning) << ex.what();
-							continue;
-						}
```
pjanck, line 158: Because you've reverted the "merge" commit, this would get deleted. We don't want that.

pjanck:
This pull request does nothing. I'll close it.

PR #186, opened 2020-07-20
Fix issue 185 with workaround

### LINKED ISSUE #185: [BUG] Generating the documentation with doxygen causes an error
**Describe the bug**
Generating the documentation with doxygen causes the following output in the error list of Visual Studio:
errorListVS

After the error happened, the dot-files are missing in the mentioned folder - one as example:
folderContent

At the moment, I can't say whether real cause of the error is an issue on my machine / settings or whether it's a general error in doxygen or dot.

**To Reproduce**
Usually, the error happens after the re-generation of the documentation multiple times, in example when I made small changes between each generation. However, after first start via CMake (including 'configure', 'generate' and 'open project') it works sometimes fine but cause the error at a second generation - but sometimes the first generation failed as well. Until now, I can't see a pattern when it works.

Steps to reproduce the behaviour:
1. In Visual Studio, 'Build' the project `OpenInfraPlatform.GenerateDocumentation` in the folder `OpenInfraPlatform/Commands/`.
2. Make changes in the code, e.g. add a new doxygen-comment.
3. 'Build' again.

**Workaround**
In the web I found the following thread, which might be a similar problem - or at least the same error message: [http://doxygen.10944.n7.nabble.com/Problems-running-dot-td653.html](http://doxygen.10944.n7.nabble.com/Problems-running-dot-td653.html)

According this thread I added the following setting in `ConfigureDoxygen.cmake`:
workAroundChanges

With this additional setting the generation works fine each time.

christophKaiser:
For description of the error and the workaround see issue #185

At the moment, I can't say whether the real cause of the error is an issue on my machine / settings or whether it's a general error in doxygen or dot.
Alternative to the bound setting of the current pull request, this setting can be provided as option in CMake. In this way, everyone with this error is able to set the option to 'no' and everyone without this error can let the option on 'yes'. ('yes' is the default if nothing else is in the config file - which is the current situation.)

Fixes #185 .

### DIFF
--- a/cmake/ConfigureDoxygen.cmake
+++ b/cmake/ConfigureDoxygen.cmake
@@ -44,6 +44,7 @@ set(DOXYGEN_PROJECT_BRIEF "Open source software for viewing IFC and point cloud 
 set(DOXYGEN_PROJECT_LOGO ${CMAKE_CURRENT_SOURCE_DIR}/UserInterface/Resources/icons/icon_72x72.png)
 set(DOXYGEN_CREATE_SUBDIRS YES)
 set(DOXYGEN_HAVE_DOT YES) # GraphViz Package for diagrams
+set(DOXYGEN_DOT_CLEANUP NO)
 set(DOXYGEN_CALL_GRAPH YES) 
 set(DOXYGEN_CALLER_GRAPH YES)
 


christophKaiser:
Maybe, the issue #185 should be linked in the right column?

pjanck:
> 
> 
> Maybe, the issue #185 should be linked in the right column?

I've edited the description of the pull request.

pjanck (commented):
LFTM. @ccoetzee Please have a look.

pjanck (approved):

ghost (approved):

PR #188, opened 2020-07-22
[REF] GeomUtils.cpp

SamuilsRulovs:
* Added exceptions to GeomUtils.cpp


pjanck (changes requested):
Good proposals. The Inconsistent..Exceptions take `string` only as input as well. Since you do not have `EXPRESSReferences` here, you should populate the exceptions with only string messages.

Core/src/IfcGeometryConverter/GeomUtils.cpp
```diff
@@ -130,23 +130,23 @@ void GeomUtils::extrude(
...
 		*/
 	}
 	if( poly_data->points.size() > 0 )
 	{
 		std::cout << "extrude: points vec should be empty" << std::endl;
 		return;
 		/*
-		throw oip::InconsistentModellingException(poly_data, "points vec should be empty");
+		throw oip::InconsistentModellingException("points vec should be empty");
 		*/
```
pjanck: All throwing code is still commented out, and the legacy code wasn't replaced.

pjanck (changes requested):
Apply similar changes to all parts, as noted with one. Replace legacy `std::cout` printing with `throw`s.

Core/src/IfcGeometryConverter/GeomUtils.cpp
SamuilsRulovs: Resolved: 3353249

Core/src/IfcGeometryConverter/GeomUtils.cpp
```diff
@@ -127,27 +127,15 @@ void GeomUtils::extrude(
 {
 	if( face_loops_input.size() == 0 )
 	{
-		std::cout << "extrude: face_loops_input.size() == 0" << std::endl;
-		return;
-		/*
-		throw oip::InconsistentModellingException("face_loops_input.size() == 0");
-		*/
+		throw oip::InconsistentModellingException("face_loops_input.size() == 0");
```
pjanck: Please include the name of the function in the messages (similarly to how it was):
```suggestion
		throw oip::InconsistentModellingException("GeomUtils::extrude(): face_loops_input.size() == 0");	
```

pjanck: Also, similarly to other messages, write out, that `face_loops_input.size() should not be 0`.

pjanck (changes requested):
Please apply the comment for one line at other lines as well.

Core/src/IfcGeometryConverter/GeomUtils.cpp
SamuilsRulovs: Resolved:  8eb52be

pjanck (approved):

pjanck (approved):

PR #201, opened 2020-08-04
Correcting EXPRESS's .cpp

### LINKED ISSUE #200: [REF] EXPRESSOptional.cpp and -Reference.cpp
**Files which are being refactored**
- `EXPRESSOptional.cpp`
- `EXPRESSReference.cpp`

The files should be header files since they define template functions. Include the functions they introduce to the corresponding header files.

pjanck:
Fixes #200 .


pjanck:
Did not solve the problem. Will submit a newer one.

PR #202, opened 2020-08-04
B-Spline - Curvature

christophKaiser:
### What's done until now
Changes in current code of `SplineConverter.h`
- In commit 76041c3, the loading process of B-Spline control points was moved into a separate function.
- In commits 6bbf52f and 248cd62, the calculation of one single curve point was moved into a separate function.
- In commits 3626bff and 95a3ecb, the calculation of the basis functions was spited into smaller parts, which will be needed for the calculation of the derivatives.

New functionality in `SplineConverter.h`
- Commits 83ae485 and 9fd6590 added functions to calculate the first and second derivative of a B-Spline.
- Commits c8ff579 and 8248ab2 add a main function `computeCurvatureOfBSplineCurveWithKnots` which calculates the curvature of a B-Spline. As well, the corresponding curve length is calculated, which is necessary to display or analyse the curvature in dependency to the true curve length. The parameter t along a B-Spline curve with open uniform knot vector doesn't represent the true length along the curve because the calculated curve points aren't evenly spaced. The function `computeCurvatureOfBSplineCurveWithKnots` is independent from an ifc entity because upcoming functions will have a `std::vector<carve::geom::vector<3>>` variable, which will be given to `computeCurvatureOfBSplineCurveWithKnots`.

### Pending work
- [done] The function `computeCurvatureOfIfcBSplineCurveWithKnots` has to be called from somewhere in the project, where the vertices in the `std::vector<carve::geom::vector<3>>` have to be obtained.
- [done] Its result, the curvature, must be interpreted.
- The final result should be displayed on the viewport.

[Update from 2020-08-20 see [below](https://github.com/tumcms/Open-Infra-Platform/pull/202#issuecomment-677907361)]
[Update from 2020-08-30 see [below](https://github.com/tumcms/Open-Infra-Platform/pull/202#issuecomment-683414473)]
[Update from 2020-11-23 see [below](https://github.com/tumcms/Open-Infra-Platform/pull/202#issuecomment-732456798)]


Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -143,7 +144,110 @@ namespace OpenInfraPlatform
...
+					 * Both return vectors have the same size. They come back as a \c std::tuple in the order \c length, \c curvature.
+					 *
+					 * \param[in]	order			Order of the B-Spline or rather the basis functions ( =degree+1 )
+					 * \param[in]	controlPoints	The vector of the B-Spline control points.
+					 * \param[in]	knotArray		The array / vector of knots.
+					 *
+					 * \return		Vector with curve length
+					 * \return		Vector with curve curvature
+					 */
+					std::tuple<std::vector<double>, std::vector<double>> computeCurvatureOfBSplineCurveWithKnots(
```
pjanck: My suggestion would be to rather use `std::vector<std::pair<double, double>>` if you're having pairs with "length along the curve" and its corresponding "curvature at that point". Access the member by `.first` and `.second` .

christophKaiser: Changed in commit 637e6d8

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -198,9 +197,7 @@ namespace OpenInfraPlatform
 						double step;
 						std::tie(knotStart, knotEnd, step) = obtainKnotRange(order, knotArray, numCurvePoints);
 
-						std::vector<double> length;
-						length.reserve(numCurvePoints);
-						std::vector<double> curvature;
+						std::vector<std::pair<double, double>> curvature;
```
pjanck: As the naming goes, `lengthsWithCurvatures` would be more descriptive, as there are not only curvatures in the vector.

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -211,28 +208,33 @@ namespace OpenInfraPlatform
...
 						for (size_t i = 0; i < numCurvePoints; ++i) {
 							if (i == numCurvePoints - 1) { t = knotEnd - accuracy; }
 
 							curvePoint = computePointOfBSpline(order, t, controlPoints, numControlPoints, knotArray);
 							if (i == 0)
-								length[0] = 0;
+								length_i = 0;
 							else
-								length[i] = length[i-1] + (curvePoint - curvePointPrevious).length();
+								length_i = curvature[i-1].first + (curvePoint - curvePointPrevious).length();
```
pjanck: A more concise `C`-style-like expression would be:

    length_i = (i == 0 ? 0 : curvature[i] ... );

pjanck (changes requested):

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: Done in commit b7aa29d.

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: Done in commit b7aa29d.
Because length_i is reduced to one line / statement, change of commit 9860558 is a suggestion from me; however, I'm not sure, whether it might got to complicated to read and understand.

pjanck: With the addition of the comments, it is transparent what is happening in the lines.

pjanck (commented):
This is starting to take on a really good shape! Good work!

Suggestion: since #220 , nice functionality for automatic testing is available. Perhaps that can help you test multiple scenarios at once.

christophKaiser:
What's happened since last description - only relevant / big changes are listed:
(Summary of the commits and one question in 'Pending Work - Now')

### User Interface
With commit bd7e81d, the new functions can be called via the UI: In the menu 'Tools' is a new option 'Convert Stroke to Alignment' (design name - maybe it will change in future).

### New class `SplineInterpretation`
- Commit 81680eb included the [existing] folder `\Rendering\Tools` into the project, where the new class was created (commit 5cfdb97).
- At this place, the main function convertSketchToAlignment calls the function computeCurvatureOfBSplineCurveWithKnots in SplineUtilities - in SplineConverter the junk function, which called the function, isn't any more needed.
- Tow functions `debugFunction_...` are available to print out some information in the console window. To see all lines, you have to change manually the property Layout > Screen Buffer Size > Height from 25 to something greater than 200. The output can be pasted into MATLAB.

### New class `SplineUtilities`
- With commits 724c5ad and d6e88b7, the new class was introduced in the folder `IfcGeometryConverter`. Some functions from SplineConverter were moved into this new class because the functions have to be called without the converter environment.

### Pending Work - Now
- [answered and done] `SplineInterpretation` should get functions to save and load the information of a `std::vector<carve::geom::vector<3>>` into a file. I prefer a simple syntax inside the save file and write an appropriate reader / writer with the class `fstream` (or `ofstream` and `ifstream`) - reading detail in the documentation is pending. At this point my question: Has to be the syntax a specific one - for example from an external software which generates test data, or the appropriate data type - or am I free in my choose?
- `SplineInterpretation` will get further functions to interpret the curvature.

### Pending Work - In (far) future
- SplineInterpretation doesn't appear in doxygen [not urgent]
- In SplineUtilities: Further investigation to include all really needed files / libraries might be reasonable. (in more detail: commit f13a07d) [not urgent]

[Update from 2020-08-30 see [below](https://github.com/tumcms/Open-Infra-Platform/pull/202#issuecomment-683414473)]

christophKaiser:
What's happened since last description

### Save and Load were added to `SplineInterpretation`

In commit 62e1341, the functions `saveControlPointsIntoFile` and `loadControlPointsFromFile` were added to `SplineInterpretation`. They work in the directory '[build folde]/testdata/StrokeToAlignment/', which has to be created via build of the project 'Copy/CopyOpenInfraPlatformUIResources' in Visual Studio.

One example file is included in the commit. The file is saved as .sketch, but can be opened with every usual text editor. The information is saved as text in syntax of a MATLAB matrix; appropriate, the matrix can be pasted into MATLAB (e.g. for debug or to visualise).

### Pending work - Now
- `SplineInterpretation` will get further functions to interpret the curvature and to create alignment objects.
- Visualisation of the alignment objects in the viewport.

### Pending Work - In (far) future
- In the code of commit 62e1341, two exception handling are missing; however, in this stage everything works fine if no exception case is triggered. In more detail, search for `// ToDo: exception handling` in the code. [not urgent]

[Update from 2020-11-23 see [below](https://github.com/tumcms/Open-Infra-Platform/pull/202#issuecomment-732456798)]

pjanck (commented):
This looks very promising. I commend you for the abundance of comments in the code!

pjanck (commented):
I'm interested to see the results of this code.

christophKaiser:
The calculations of curvature and curvature change were tested with the following curve:
0 curve
The points are a perfect alignment obtained by CAD.

In MATLAB, I have almost the similar functions to compute the B-Spline, its curvature with change of curvature, and smoothing via moving average & variogram variable. The following two charts were my result, one or two weeks ago:
1 expected curvature
2 expected curvatureChange
The first one shows the calculated curvature (blue line) with the smoothed data (orange line). The second one shows the change of curvature obtained by numerical derivative (blue line) and after smoothing (orange line).
In summary, the result was very good at this point.

The implementation in OIP is almost like in MATLAB; thus, I expected the same results. However, I got the following one (visualization with MATLAB, but calculation with OIP):
3 real curvature
4 real curvatureChange
As well, the figures show the curvature and the change of curvature. The calculation is displayed in grey, the result of smoothing in black. In total, this result is very bad because the curvature doesn't show the straight section and the change of curvature doesn't say anything useful.

The reason for this result is a very high value of range in the variogram, which is used as window size of the moving average. The low value in MATLAB might be caused by the peaks to +3.2 and -3.5 near the start and end of the curvature plot; in the B-Spline calculation of OIP, this peak doesn't appear.
The high value of range in OIP isn't caused by a mistake in the implementation. By calling my smoothing function in MATLAB with the calculated curvature data from OIP, MATLAB had the same result.

Now with the limit of range in commit fcd1513, the charts are the following:
5 afterFix curvature
6 afterFix curvatureChange
The smoothing of perfect data works better.

As well, I have an example of a sketched alignment:
7 sketch curve
8 sketch curvature
9 sketch curvatureChange
The quality of smoothing is okay, especially in the range of curve length 0 to 400 short disturbances in curvature are smoothed well.

pjanck:
This looks very nice. The graphs of the "real world" example seem very wobbly, I'm not sure if you will even recognize the straight section in the middle. However, the synthetic results look very good. 

A small hint regarding visualization: it is suboptimal to use grey and black - when copying this information can get hazy. It is better to use different line patterns, like full line, dashed line, dotted line, dash-dot etc.

Rendering/Tools/SplineInterpretation.cpp
```diff
@@ -345,6 +370,185 @@ std::vector<double> OpenInfraPlatform::UserInterface::SplineInterpretation::nume
...
+			indicator[i] = 0;
+		else
+			indicator[i] = 1;
+
+	return { indicator, curvatureZero };
+}
+
+std::vector<SplineInterpretationElement> OpenInfraPlatform::UserInterface::SplineInterpretation::obtainElementsFromIndicator(
+	std::vector<int> indicator,
+	std::vector<std::pair<double, double>> lengthsWithCurvatures) const throw(...)
```
pjanck: if you are not changing a variable within the function, it is advisable to pass it as `const &` - saves the copying time for the processor.

```suggestion
	const std::vector<int>& indicator,
	const std::vector<std::pair<double, double>>& lengthsWithCurvatures) const throw(...)
```

Primitive types (like `int` and `double`) don't need the reference `&`.

Rendering/Tools/SplineInterpretationElement.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+	length = _length;
+	indicator = _indicator;
+	indices = _indices;
+}
+
+
+// PUBLIC FUNCTIONS WITH ACCESS TO VARIABLES
+void OpenInfraPlatform::UserInterface::SplineInterpretationElement::setLength(double newLength) throw(...){
+	length = newLength; }
```
pjanck: Such a function does not throw, there is no `throw` statement within it. Use `throw()` to denote that.

pjanck (commented):
Small suggestions for code improvement.

Rendering/Tools/SplineInterpretationElement.cpp
christophKaiser: In commit b492fab, I removed the `throw` from the get- and set-functions.
Do you mean with "`throw` statement" something like the following lines?
```
if(badCase)
   throw oip::UnhandledException(myVariable);
```

pjanck: Yes, sth like this. Beware, though: if you use `throw()` as an indicator to a function definition, then neither the function itself or the functions it calls are not allowed to throw. Which means - if you are not throwing anything in the function and you do not call another function, where you are unsure if it throws, you can use the `throw()` indicator.

christophKaiser:
Now, there are results from the data of the charts which I showed you a few days ago ([link to comment](https://github.com/tumcms/Open-Infra-Platform/pull/202#issuecomment-703169881)).

First the results to the calculated data:
01 KreisKlothoideGerade Berechnet Ergebnis

There is a drawing (with AutoCAD) which shows the resulting elements from the calculated data in orange. The element endpoints are highlighted with a circle point. The ends of the clothoids don't match perfect, maybe it's an issue of accuracy in my Excel calculation of the clothoid curve points (support tool to draw clothoids in AutoCAD).
Under the orange line is a white line which shows the original alignment.
01 KreisKlothoideGerade Berechnet Ergebnis Zeich

Here are the results from the sketched data:
01 KreisKlothoideGerade Skizze Ergebnis

christophKaiser:
The last commit 917a313 is the source code at end of the BA.

christophKaiser:
The following listing is an overview of pending work for optimisation, which was collected in the above comments and commit messages. The order of my tasks represents more or less the priority according to my opinion. However, it’s free for discussion.
- [ready for review, commits 6bab0e3 and a93fbd7] The class `SplineInterpretation` doesn't appear in doxygen (help wanted)
- [ready for review, commit 2feabf6] Complete doxygen comments in `SplineInterpretation.h`.
- [ready for review, commit 07889a5] In `SplineUtilities`: Further investigation to include all really needed files / libraries might be reasonable. (in more detail: commit f13a07d)
- [canceld, get wrong information] Modify function definitions in cpp-files by use of `namespace` or `using namespace`. This should short the definitions for more readability. (Experience from current CiE 1 course.)
- [ready for review, commit 51a8d46] Add exception handling if loading of a file fails. (`SplineInterpretation::loadControlPointsFromFile()`, line 208)
- [ready for review, commit ccc183c] Add exception handling if `controlPoints` is an empty vector after loading. (`SplineInterpretation::convertSketchToAlignment()`, line 44)
- [ready for review, commit 154ca21] Rework `const` and `throw`. (regarding to pjanck's comment [above](https://github.com/tumcms/Open-Infra-Platform/pull/202#pullrequestreview-501728744))
- The final result (input points and alignment elements) should be displayed on the viewport.

Optional (currently not planed)
- Performance work: The function `convertSketchToAlignment()` needs about 2 minutes for the result in my case and with 168 control points (example `001 Testsketch ArcClothoidStraight calculated.sketch`). From this time, the function `splineUtilities.computeCurvatureOfBSplineCurveWithKnots(…)` needs about 1:40 minutes. At this point, further observation and analysis of the code is necessary. Maybe, the calculation of the derivatives and their basis functions must be redesigned.
- Create a unit test, if displaying on the viewport was implemented successful.
- Implementation of sketch input.

EDIT 7 Jan 2021: Added progress in square brackets.
EDIT 8 Jan 2021: 'Performance work' was moved to optional (not planed) (according to today's meeting)
EDIT 26 Feb 2021: Updated progress in square brackets.

christophKaiser:
Updated the comment with pending work from 23 Nov 2020 (see [here](https://github.com/tumcms/Open-Infra-Platform/pull/202#issuecomment-732456798)).
Additional commits from above:
- Small correction and comment optimation in commits 97447bf and 3ae2877.
- Additional message box to inform about finished calculation in commit 9988bca.

Rendering/Tools/SplineInterpretation.cpp
```diff
@@ -15,6 +15,8 @@
 	along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
 
+#pragma once
+
```
pjanck: No need for this - this is useful in headers.

Rendering/Tools/SplineInterpretation.h
```diff
@@ -121,6 +124,14 @@ namespace OpenInfraPlatform
...
 
+			/* \brief In a vector of pairs, this function obtains the second pair value and saves the values into a new vector
+			 *
+			 * For example the curvature in the vector \c lengthsWithCurvatures can be obtained for further use.
+			 *
+			 * \param[in]	lengthsWithCurvatures	A vecotr of \c pair<double,double>.
+			 *
+			 * \return		A vector of \c double wiht the second values of the pair vector.
+			 */
 			std::vector<double> obtainCurvatureFrom_lengthWithCurvatures(const std::vector<std::pair<double, double>>& lengthsWithCurvatures) const throw(...);
```
pjanck: This is easily obtained with a `std::transform` locally.

pjanck (commented):
Thank you for the doxy-comments - now it is much clearer, what the individual functions do and produce.

Rendering/Tools/SplineInterpretation.cpp
christophKaiser: Reverted in commit 00bf496.

Rendering/Tools/SplineInterpretation.h
christophKaiser: The function calls are replaced to `std::tansform` in commit 75dcd42. As well, one function overload of obtainThreshold is removed entirely, instead `std::transform` is used at its previous call.

christophKaiser:
**About 'The class `SplineInterpretation` doesn't appear in doxygen'** (see pending work [here](https://github.com/tumcms/Open-Infra-Platform/pull/202#issuecomment-732456798))

The source files of the classes `SplineInterpretation` and `SplineInterpetationElement` are stored in the directory `/Rendering/Tools/`.

With the following line in the general Doxygen settings all content in folders 'Tools' is excluded (besides other folders), what is the reason why the classes doesn't appear in the generated documentation.
https://github.com/tumcms/Open-Infra-Platform/blob/90e74889a85fb7d1044419a0297c657c95d38355/cmake/ConfigureDoxygen.cmake#L76

If I remove the item `*/Tools/*` from this list (immediately before the last item), the mentioned classes appear in the documentation. However, I'm not sure whether this is the best solutuion because at somepoint in past someone (probably) had a reason to exclude that folders (there are other Tools folders in other subdirectories of the project).

An alternative solution might be to move these source files into another folder which isn't excluded from the documentation - or just rename the appropriate folder to fall out of the exclude pattern.

Probably, there might be another solution as well which I don't know. @pjanck  At this point, I want to ask for help to get / found the best solution for that issue - especially in context and with respect to the hole project.

pjanck:
I propose to move your content to `Core/SplineInterpretation` directory.

Be sure to amend `Core/CMakeLists.txt` to include the files in the solution when generating with CMake.

Core/CMakeLists.txt
```diff
@@ -28,13 +28,15 @@ file(GLOB Core_DataManagement_General_Source 		src/DataManagement/General/*.*)
 file(GLOB Core_IfcGeometryConverter_Source			src/IfcGeometryConverter/*.*)
 file(GLOB Core_OffConverter_Source					src/OffConverter/*.*)
 file(GLOB Core_Exception_Source						src/Exception/*.*)
+file(GLOB Core_SplineInterpretation_Scource			src/SplineInterpretation/*.*)
```
pjanck: Typo - a "c" too many :)

Core/src/SplineInterpretation/SplineInterpretation.cpp
```diff
@@ -454,20 +444,14 @@ std::tuple<std::vector<int>, double> OpenInfraPlatform::UserInterface::SplineInt
...
 }
 
-double OpenInfraPlatform::UserInterface::SplineInterpretation::obtainThreshold(
-	const std::vector<std::pair<double, double>>& lengthsWithCurvatures) const throw(...)
-{
-	return obtainThreshold(obtainCurvatureFrom_lengthWithCurvatures(lengthsWithCurvatures));
-}
-
-double OpenInfraPlatform::UserInterface::SplineInterpretation::obtainThreshold(std::vector<double> data) const throw(...) {
+double OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretation::obtainThreshold(std::vector<double> data) const throw(...) {
```
pjanck: Beginning with vs2019 (which we'll be switching to soon) the `throw(...)` should be replaced with `noexcept(false)`. Please replace in all new files written by you accordingly.

pjanck (changes requested):
Some small suggestions.

Core/CMakeLists.txt
christophKaiser: Done in commit 347f2ab

Core/src/SplineInterpretation/SplineInterpretation.cpp
christophKaiser: Done in commits 94afcd3 and 615d2b1

Rendering/Tools/SplineInterpretation.cpp
christophKaiser: Done commit 154ca21

Core/src/SplineInterpretation/SplineInterpretation.cpp
```diff
@@ -263,15 +263,15 @@ std::vector<carve::geom::vector<3>> OpenInfraPlatform::Core::SplineInterpretatio
 
 		return points;
 	}
-	catch (const std::exception& e)
+	catch (const std::exception&)
 	{
 		// in case of an error, return an empty vector
 		std::vector<carve::geom::vector<3>> pointsEmpty;
 		return pointsEmpty;
 	}
```
pjanck: It is advisable to at least log that an error occured. Use sth similiar to:

    BLUE_LOG(warning) << "error occured ... ";

Be as specific as possible, so that the future programmer knows where to look.

Core/src/SplineInterpretation/SplineInterpretation.cpp
```diff
@@ -368,10 +368,11 @@ size_t OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretation::vari
 		gamma[h - 1] /= (2 * N);
```
pjanck:
```suggestion
		gamma[h - 1] /= (2. * N);
```

It is advisable to use floating numbers when dividing, so that the result is not rounded.

Core/src/SplineInterpretation/SplineInterpretation.cpp
```diff
@@ -771,16 +768,19 @@ SplineInterpretationElement OpenInfraPlatform::Core::SplineInterpretation::Splin
...
 {
 	element.setType("clothoid");
 	element.setStartpoint(startPoint);
 
 	// radius of clothoid (for calculation)
-	double radius;
+	//   0 is just a preset value;
+	//   the actual one will be obtained surely in one of the following if-cases,
+	//   because of the requirement of one arc at one clothoid end
+	double radius = 0;
```
pjanck:
```suggestion
	double radius = 0.;
```

Initialize floating variables with floating numbers. Always.

Core/src/SplineInterpretation/SplineInterpretation.cpp
```diff
@@ -862,12 +862,12 @@ carve::geom::vector<3>  OpenInfraPlatform::Core::SplineInterpretation::SplineInt
 }
 
 carve::geom::vector<3> OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretation::tangentVectorFromDirection(
-	double direction) const throw(...)
+	const double direction) const noexcept(true)
 {
 	return carve::geom::VECTOR(cos(direction*M_PI / 180), sin(direction*M_PI / 180), 0.0);
```
pjanck:
```suggestion
	return carve::geom::VECTOR(cos(direction*M_PI / 180.), sin(direction*M_PI / 180.), 0.0);
```

How come that direction is in degrees? Naturally, in C, angles are in radians. Where does this come from?

pjanck (changes requested):
Small suggestions. Should be applied everywhere ( I'm not sure I managed to get all occurrences).

Core/src/SplineInterpretation/SplineInterpretation.cpp
christophKaiser: Done in commit 643376a.

Core/src/SplineInterpretation/SplineInterpretation.cpp
christophKaiser: Done in commit e4fc298.

Core/src/SplineInterpretation/SplineInterpretation.cpp
christophKaiser: Done in commit e4fc298.

Core/src/SplineInterpretation/SplineInterpretation.cpp
christophKaiser: Done in commit 3ca85d7.
These angles are calculated in the function `angleOfVectors2D` where they were converted to degree manually.
During the development process, I was focused on testing with CAD drawings where usually degree is used. According to https://standards.buildingsmart.org, both degree and radiant is possible in an ifc file. At this point, I probably missed to doublecheck what is used in the ifc example files.
With the current changes, everything program internal works in radian and just the print output to the console window is in radian and degree (just in case that the value in degree might still be usefull in some further usage or human interpretation).

pjanck: Right, I understand. Rule of thumb:

- Internally, OIP works in meters and radians. 
- Whenever we print, we may decide and reformat to any other unit.

Core/src/SplineInterpretation/SplineInterpretationElement.cpp
```diff
@@ -107,6 +107,9 @@ void OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement:
 double OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement::getDirection() const noexcept(true) {
 	return direction_; }
 
+double OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement::getDircetionDegree() const noexcept(true) {
```
pjanck: typo in function name

Core/src/SplineInterpretation/SplineInterpretationElement.cpp
```diff
@@ -128,6 +131,9 @@ void OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement:
 double OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement::getAngle() const noexcept(true) {
 	return angle_; }
 
+double OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement::getAngleDegree() const noexcept(true) {
+	return angle_ * 180.0 / M_PI; }
+
```
pjanck:
```suggestion
double OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement::getAngle( bool inDegrees = false ) const noexcept(true) {
	return angle_ * (inDegrees ? 180.0 / M_PI : 1.) ; }
```

Function with a default parameter could be considered. (syntax may be wrong, but I believe you get the idea)

Core/src/SplineInterpretation/SplineInterpretationElement.cpp
```diff
@@ -128,6 +131,9 @@ void OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement:
 double OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement::getAngle() const noexcept(true) {
 	return angle_; }
 
+double OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement::getAngleDegree() const noexcept(true) {
+	return angle_ * 180.0 / M_PI; }
+
 
 void OpenInfraPlatform::Core::SplineInterpretation::SplineInterpretationElement::setIsCCW(const int isCCW) noexcept(true) {
```
pjanck, line 132: how come `isCCW` is an `int` and not a `bool`?

pjanck (changes requested):

Core/src/SplineInterpretation/SplineInterpretationElement.cpp
christophKaiser: Function was removed by changes according to the suggestion of using default parameter (see next below).
Done in commit d8bb82e.

Core/src/SplineInterpretation/SplineInterpretationElement.cpp
christophKaiser: Thanks for that suggestion! The concept of default parameter was new to me.
Done in commit d8bb82e.

Core/src/SplineInterpretation/SplineInterpretationElement.cpp
christophKaiser: To be honest, I can't remember.
It's changed to `bool` in commit 1d19e35.

christophKaiser:
Additionally, to the comments in the conversations, commit f5f402c fixes a few syntax mistakes in the Doxygen comments. Probably I didn't recognise them before because they are in private functions which don't appear in the documentation by default.

pjanck (approved):
I think that this is now done.

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -565,6 +364,8 @@ namespace OpenInfraPlatform
 
 					std::shared_ptr<PlacementConverterT<IfcEntityTypesT>> placementConverter;
 
+					SplineUtilities splineUtilities;
+
```
pjanck: I'd propose to make the functions `static` and omit the need for a member here.

Core/src/IfcGeometryConverter/SplineUtilities.h
```diff
@@ -0,0 +1,768 @@
...
+				 *
+				 * \return		The array of curve points, which can be rendered in a viewport after correction by \c GeomUtils::appendPointsToCurve.
+				 */
+				 // B-Spline curve definition according to: http://mathworld.wolfram.com/B-Spline.html
+				std::vector<carve::geom::vector<3>> computeBSplineCurveWithKnots(
+					const int& order,
+					const std::vector<double>& knotArray,
+					const std::vector<carve::geom::vector<3>>& controlPoints,
+					const uint32_t& numCurvePoints,
+					const double& accuracy) const throw(...)
```
pjanck: For example, this can definitely be a `static` function.

pjanck (approved):
I say we merge this and move forward with implementing B-Spline curves from IFC (after addressing the proposals below).

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: Done in commit 4f9851b.
Now, all functions of SplineUtilities are static because they don't store information in member variables of SplineUtilities. As well, these functions don't call other non-static functions, thus your suggestion is comprehensible - thanks.

Core/src/IfcGeometryConverter/SplineUtilities.h
christophKaiser: Done in commit 4f9851b.

pjanck (approved):

christophKaiser:
Commit cb3a5a4 addes files to do a [manual] unit test of the function 'Core::SplineInterpretation::SplineInterpretation::convertSketchToAlignment' including its subfunctions. Commit ebfac5a provides a optimization of the print output in the console window.
Issues about the pending work (which aren't urgent) are created (#408 #409 #410 #411 )- according to the last dev meeting.

The current commit of the branch `tumcms/development` was merged into this branch with commit c4da9bd.
From my side, this pull request is ready to be merged into the devolpment branch.

pjanck (approved):

christophKaiser:
I forgot to mention: The provided example files of Stroke to Alignment will become available after build of the project */Copy/CopyOpenInfraPlatformUIResources*:
BuildProject

Once built, you can start the function via the menu *Tools* > *Convert Stroke to Alignment*. After chooseing of the option *Load*, you should get the following open dialog which shows you three files in the build directory:
FilesInOpendialog

You can store additional sample files in this folder, without effects on the directory which is tracked by git.

PR #209, opened 2020-08-06
Visual unit tests

hechth:
Implemented basic version of visual unit tests so that it can be developed further in the main branch.


pjanck (approved):
LFTM.

PR #210, opened 2020-08-07
Only entity attributes

### LINKED ISSUE #203: [REF] Use base class functions for common attribute functionality
**File which is being refactored**
GeneratorOIP.cpp

**Functions**
- [ ] generateEntitySourceFileREFACTORED
- [ ] generateEntityHEaderFileREFACTORED

Writing of swap function, assignment operator, ReadStepData, GetStepLine has to be refactored in order to use base class functions so that not all attributes have to be reference or therefore included in the class.

pjanck:
Eventually fixes #203 .

Currently does not fix anything - just shuffles some code around. Classes call the corresponding functions of their parent classes (entities). Should consider changing the types as well.

Reading of an IFC file is not working currently. I assume it has to do with the interpretation of references.

### DIFF
--- a/EarlyBinding/src/EXPRESS/EXPRESSEntity.h
+++ b/EarlyBinding/src/EXPRESS/EXPRESSEntity.h
@@ -27,10 +27,14 @@
 
 #include <tuple>
 #include <utility>
+#include <algorithm>
+#include <vector>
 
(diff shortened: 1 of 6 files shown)


pjanck:
This does not bring the expected benefits. Will not continue to work on this.

PR #211, opened 2020-08-07
Refactoring ParamterizedProfileDef

SamuilsRulovs:
Refactored "convertIfcParameterizedProfileDef" so "convertIfcParameterizedProfileDefWithPosition" would be its sub-function


pjanck (approved):
Nicely done. This seems correct.

We need to test this with example files from the official web-documentation.

Core/src/IfcGeometryConverter/ProfileConverter.h
```diff
@@ -1610,6 +1545,35 @@ namespace OpenInfraPlatform {
...
+				if (profileDef->Position) {
+					carve::math::Matrix transform = placementConverter->convertIfcPlacement(profileDef->Position.get());
+
+					for (std::vector<carve::geom::vector<2>> path_loop : temp_paths) {
+						for (carve::geom::vector<2> pt : path_loop) {
+							carve::geom::vector<3> pt_3d(carve::geom::VECTOR(pt.x, pt.y, 0));
+							pt_3d = transform * pt_3d;
+							pt.x = pt_3d.x;
+							pt.y = pt_3d.y;
+						}
```
pjanck: Are you trying to overwrite `pt` here? Was it like this also before the changes?

Becase then you would need to initialize the running variable with `&` as: `carve::geom::vector<2> &pt : path_loop`.

EDIT: Yes, it was with `&` before the change. Please be careful with such things.

pjanck (changes requested):
There seem to be a mistake in applying the transformation from the position.

Core/src/IfcGeometryConverter/ProfileConverter.h
SamuilsRulovs: Resolved: 5ef7ef1

pjanck (approved):

PR #212, opened 2020-08-07
Unit test utilities

hechth:
* Implemented unit test utility for rendering of IfcGeometryModel.
* Updated to new BlueFramework with improved camera controls.
* Renderer is now able to change the viewing angle of the model.


CMakeLists.txt
```diff
@@ -236,7 +236,7 @@ endif()
 FetchContent_Declare(
 	blue_framework
 	GIT_REPOSITORY https://github.com/tumcms/Blue-Framework.git
-	GIT_TAG a0.0.7
+	GIT_TAG a0.0.8
```
pjanck, line 239: Does CMake download the update automatically when generating, or does one need to clean cache, delete binaries and set up anew?

pjanck (approved):
Exemplary work!

PR #215, opened 2020-08-12
Created Unite Tests for extruded-solid and surface-model

SamuilsRulovs:
Have created Unit Test for:
extruded-solid.ifc
surface-model.ifc


UnitTests/Schemas/IFC4X1/extruded-solid/src/extruded-solid.cpp
```diff
@@ -107,7 +63,7 @@ class ExtrudedSolidTest : public VisualTest {
 	void compareImageWithView(const std::string filename) {
 
 	// Arrange
-	const auto expected = buw::loadImage4b(boost::dll::program_location().parent_path().concat(filename).string());
+	const auto expected = buw::loadImage4b(testPath(filename).string());
```
pjanck: This should probably be `dataPath` - you are loading a saved picture, right?

UnitTests/Schemas/IFC4X1/extruded-solid/src/extruded-solid.cpp
```diff
@@ -173,32 +129,32 @@ TEST_F(ExtrudedSolidTest, BottomView)
...
+	buw::storeImage(testPath("extruded-solid_bottom.png").string(), image);
 
 	// Assert
 	EXPECT_NE(image, _background);
 }
 
 TEST_F(ExtrudedSolidTest, GivenNewImage_AfterHome_AreEqual)
 {
-	compareImageWithView("\\extruded-solid.png");
+	compareImageWithView("extruded-solid.png");
```
pjanck: Although I like the idea of having a dedicated function for this, it is not common practice. Each Test case should be independent of the others.

UnitTests/Schemas/IFC4X1/surface-model/src/surface-model.cpp
```diff
@@ -63,7 +63,7 @@ class SurfaceModelTest : public VisualTest {
 	void compareImageWithView(const std::string filename) {
 
 	// Arrange
-	const auto expected = buw::loadImage4b(boost::dll::program_location().parent_path().concat(filename).string());
+	const auto expected = buw::loadImage4b(testPath(filename).string());
```
pjanck: Again, `dataPath` would probably be the beter choice.

pjanck (changes requested):
Although I really like the idea of having the dedicated function for comparing the numerous views, I would discourage you from using it. Each `TEST_F` should be readable for itself.

Another idea worth considering: having one `TEST_F` checking *all* views:
    
    TEST_F(.. , AllViews) {
    expected_top = loadImage( _top.png );
    expected_bottom = loadImage ( _bottom.jpg );
    etc.
    renderer->setViewDirection( ::Top );
    image_top = CaptureImage();
    etc.
    EXPECT_EQ( expected_top, image_top);
    EXPECT_EQ( expected_bottom, image_bottom);
    }

UnitTests/Schemas/IFC4X1/extruded-solid/src/extruded-solid.cpp
SamuilsRulovs: Resolved: 2b8ebfb

UnitTests/Schemas/IFC4X1/extruded-solid/src/extruded-solid.cpp
SamuilsRulovs: Resolved: 2b8ebfb

UnitTests/Schemas/IFC4X1/surface-model/src/surface-model.cpp
SamuilsRulovs: Resolved: 2b8ebfb

UnitTests/Schemas/IFC4X1/extruded-solid/src/extruded-solid.cpp
```diff
@@ -96,65 +83,202 @@ TEST_F(ExtrudedSolidTest, ImageIsSaved)
...
+	EXPECT_EQ(image_top_left_front, expected_top_left_front);
+	EXPECT_EQ(image_top_front_right, expected_top_front_right);
+	EXPECT_EQ(image_top_left_back, expected_top_left_back);
+	EXPECT_EQ(image_top_right_back, expected_top_right_back);
+	EXPECT_EQ(image_back_left_bottom, expected_back_left_bottom);
+	EXPECT_EQ(image_right_bottom_back, expected_right_bottom_back);
+
+	// Annihilate
 
+}
```
pjanck: Is this a mere duplication of the other functions? If yes, you can delete these, since you are checking the same thing twice.

pjanck (changes requested):
Similar comment applies to the other unit test as well. Otherwise: exemplary work! When corrected this pull request can be merged and you can start working on the other examples.

UnitTests/Schemas/IFC4X1/extruded-solid/src/extruded-solid.cpp
SamuilsRulovs: Resolved: 16c401a

pjanck (approved):

UnitTests/Schemas/IFC4X1/extruded-solid/src/extruded-solid.cpp
```diff
@@ -64,7 +64,12 @@ class ExtrudedSolidTest : public VisualTest {
 TEST_F(ExtrudedSolidTest, AllEntitiesAreRead) {
 	EXPECT_THAT(express_model->entities.size(), Eq(30));
 }
-
+//TODO:
+/*
+TEST_F(ExtrudedSolidTest, IFCHaveAnEssentialEntity) {
+	
+}
+*/
```
pjanck: Perfect, exactly what I had in mind :)

pjanck (approved):

UnitTests/Schemas/IFC4X1/extruded-solid/src/extruded-solid.cpp
SamuilsRulovs: Now it should also work after fa2b8df :)

pjanck (changes requested):
No need to read the file manually like that. The `IFCReader` takes care of that.

You can exchange the whole code with these lines (exchange the `...` with the correct entity class):

    auto result = std::find_if( express_model->entities.begin(), express_model->entities.end(), []( auto &pair ) -> { return pair.second.isOfType<...>(); } );
    EXPECT_NE(result, express_model->entities.end() );

*note:* perhaps the syntax is not completely correct. I'm writing this online without properly trying it out myself.

You can read more on `std::find_if` [here](http://www.cplusplus.com/reference/algorithm/find_if/).

You can read more on `gtest` [here](http://www.yolinux.com/TUTORIALS/Cpp-GoogleTest.html).

pjanck (approved):

PR #247, opened 2020-10-26
Updated ::readStepData

### LINKED ISSUE #176: [BUG] Strings still include '...' from IFC file
**Describe the bug**
String literals are read in, but still contain the `'..'` from the IFC file. That is:
- within IFC file: `'degree'`
- in VS: `"'degree'"`

**To Reproduce**
1. Import https://github.com/tumcms/Open-Infra-Platform/blob/f4558452ee7a43fd4983589bf21163cf67fe69fc/testdata/IFC4x1ExampleFiles/official/E15_Alignment/sectioned-solid.ifc
2. Breakpoint in https://github.com/tumcms/Open-Infra-Platform/blob/dac55dabc4dd0f79fcd67a10bb88520af07b720e/Core/src/IfcGeometryConverter/UnitConverter.h#L229
3. See the error, stemming from https://github.com/tumcms/Open-Infra-Platform/blob/f4558452ee7a43fd4983589bf21163cf67fe69fc/testdata/IFC4x1ExampleFiles/official/E15_Alignment/sectioned-solid.ifc#L38

**Expected behavior**
Instead of `"'degree'"` only `"degree"` would be preferred.

SamuilsRulovs:
Updated ::readStepData, so it could remove ' ' from input strings

Fixes #176

### DIFF
--- a/Core/src/IfcGeometryConverter/UnitConverter.h
+++ b/Core/src/IfcGeometryConverter/UnitConverter.h
@@ -227,7 +227,7 @@ namespace OpenInfraPlatform
 						double unitFactor = convertUnit(conversionBasedUnit->ConversionFactor->UnitComponent);
 
 						//tweak until #36 gets resolved
-						if (conversionBasedUnit->Name == "'degree'")
+						if (conversionBasedUnit->Name == "degree")
 							value = 0.017453293;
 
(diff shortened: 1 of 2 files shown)


pjanck (approved):
Seems right.

PR #330, opened 2021-01-17
Introducing IFC4x3_rc2

pjanck:
Added:
- IFC4x3_rc2 schema (stand 2020-12-13)
- CMake flags
- IfcPeekStepReader enum

Additionally:
- amended the generator a bit to account for `IfcLinearPlacementElement` - `IfcBoundedCurve` inverse - now the .Mid lib has more elements, the overall size of IFC libs with compiled objects stayed similar (approx. 17 GB)
- introduced some really bad fixes to Placement + Curve converters to get the code to compile - *this needs rework*
- there are some newly introduced geometries (curves) - *further work*


pjanck:
Replaced by #361 .

PR #368, opened 2021-02-14
Updated screenshots

Elvira2227:
screenshots of the following UniTests have been updated:
1.With the correct geometry
2.With Tests present in the code (without comments /**/)
3. No errors in compilation


pjanck (approved):
LFTM.

Although it is really interesting, as to what exactly are the differences between the previous and current versions. Perhaps @hechth can explain?

pjanck:
Additional question: do these tests succeed? That is, do they run without error?

PR #380, opened 2021-02-24
Add test files for future off unit tests

jschlenger:
Adding .off example files to the testdata folder. 
This will be the basis for a new issue regarding the development of more unit tests for the .off file format


testdata/Off/boxcube.off
```diff
@@ -0,0 +1,108 @@
+Source:
+This OFF model was downloaded from http://www.holmes3d.net/graphics/offfiles/
+and was not created by the developers of the Open Infra Platform.
+
```
pjanck: I believe that this makes it an invalid OFF file. Shouldn't this be marked as "comments" in OFF format?

pjanck (changes requested):
Apply to all files.

testdata/Off/boxcube.off
jschlenger: Resolved: c45de96

pjanck (approved):

PR #422, opened 2021-04-08
Including schemas in solution

pjanck:
Simple tweak, so that IFC schemas now also appear in the solution.

Only those schemas currently allowed to be chosen in CMake are added.


jschlenger (approved):
Looks good

PR #428, opened 2021-04-12
Unit test "terrain-surface"

### LINKED ISSUE #79: Example file "terrain-surface" is not working.
The file is [here](../blob/development/testdata/IFC4x1ExampleFiles/official/E15_Alignment/terrain-surface.ifc). Current version crashes.

Elvira2227:
Screenshots updated

Fixes #79


pjanck (changes requested):
This doesn't seem right. The position of the alignment is way off. Should be further investigated.

PR #446, opened 2021-04-25
Screen shot maker

pjanck:
Made a small console application that saves all screen shots needed for unit testing.

Usage:

> `OpenInfraPlatform.ScreenShotMaker.exe <pathToIfcFile> -o <pathToOutputDirectory>`

with:

- `<pathToIfcFile>`: the relative or absolute path to the IFC file to be rendered
- `<pathToOutputDirectory>`: the relative or absolute path to the directory where screen shots should be saved

*NOTE:* make sure that the correct IFC version is being compiled - select accordingly in CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,6 +161,18 @@ endif()
 
 
 FetchContent_Declare(
+	tclap
+	GIT_REPOSITORY https://github.com/mirror/tclap.git
+	GIT_TAG v1.2.2
+)
+  
+FetchContent_GetProperties(tclap)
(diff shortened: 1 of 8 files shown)


jschlenger (approved):
Looks good to me. 
Seems like a useful tool to simplify writing unit test.
I think it would make sense to create a new issues to change the existing unit tests to make use of this tool.

pjanck:
This is not a library, but an `.exe` - it runs independently. So there is no need for existing unit tests to be adapted.

Issue #80, opened 2020-04-23
Example file "vertical-alignment" is not working.


SamuilsRulovs:
The file is [here](../blob/development/testdata/IFC4x1ExampleFiles/official/E15_Alignment/vertical-alignment.ifc). Current version crashes.

SamuilsRulovs:
As of today, still not working

Issue #76, opened 2020-04-23
Example file "linear-placement" is not working.


SamuilsRulovs:
The file is [here](../blob/development/testdata/IFC4x1ExampleFiles/official/E15_Alignment/linear-placement.ifc). Current version crashes.

SamuilsRulovs:
As of today, still not working

Issue #94, opened 2020-04-26
Performance is lacking


pjanck:
See files #78 and #79 . It takes forever that the file is read, although there are < 200 entities. 

The main reason is handling long lists within certain entities. How can this be improved?

PR #447, opened 2021-05-07
Partly fixing #76 and #80 - Example file not working

christophKaiser:
While loading `vertical-alignment.ifc` and `linear-placement.ifc`, I got the following error, which was probably meant as 'crash' in #76 and #80 by @SamuilsRulovs:
vertical-alignment 01 errorMessage

After change the arguments of `IFCSIUNIT` from `$` to `*` (according to commit '[Corrected all IFCSIUNIT-s in all files.](https://github.com/tumcms/Open-Infra-Platform/commit/96e9eaab22a57c035f9fffb2fab135d667b73255)') the error doesn't appear anymore.

However, the vertical alignment doesn't look correct in the viewport (kink / jump at the right side of the picture):
vertical-alignment 02 wrongModelAfterFix
The linear placement loads, as well - but it takes quite long. However, I haven't had time to investigate the result in more detail.

**Thus, both issues probably wouldn't be fixed completly with this small change - but it will improve the current situation at least a bit.**

I came over that because I had the similar error message (but with different background reason) in my current issues about `IfcBSplineSurfaceWithKnots`, respectively `basin-advanced-brep.ifc`.


pjanck (approved):
These changes make sense.

pjanck:
As for the other comments:

- every file with a terrain model takes long - string parsing of the TIN is really slow (see #94 )
- the vertical calculation is somehow wrong (but only at the end of the alignment - known rpoblem, at least to me)

jschlenger (approved):
Looks fine to me

PR #342, opened 2021-01-22
Unit test "polygonal-face-tessellation"

Elvira2227:
Added new unit test "polygonal-face-tessellation"

Fixes #61 closes #317

P.S.: Please check Line#18 with describe in #338

PR #448, opened 2021-05-09
Implementing IfcIndexedPolygonalFaceWithVoids

### LINKED ISSUE #61: Example file "polygonal-face-tessellation" not rendered.
The file is [here](../blob/development/testdata/IFC4x1ExampleFiles/official/E4_Tessellated_shape/polygonal-face-tessellation.ifc).  OIP doesn't show anything.

### LINKED ISSUE #317: [BUG] Official IfcPolygonalFaceSet example is wrong
**Describe the bug**
See [here](https://forums.buildingsmart.org/t/ifcpolygonalfaceset-and-face-winding/3169/3)

**Expected behavior**
- [ ] Correct the example file, so that face winding is correct.
- [ ] Update unit tests if needed.

SamuilsRulovs:
Fixes #61 closes #317

### DIFF
--- a/Core/src/IfcGeometryConverter/FaceConverter.h
+++ b/Core/src/IfcGeometryConverter/FaceConverter.h
@@ -1255,13 +1255,14 @@ namespace OpenInfraPlatform {
 							carve::geom::vector<3> vertex =
 								carve::geom::VECTOR(point[0],
 									point[1],
-									point[2]) * length_factor;
+									point[2]) * UnitConvert()->getLengthInMeterFactor();
 
 							// apply transformation
 							vertex = pos * vertex;
 
 							polygon->addVertex(vertex);
 						}
+
 
 						auto& faces = faceSet->Faces;
 						auto& pnIndices = faceSet->PnIndex; // optional
@@ -1274,8 +1275,27 @@ namespace OpenInfraPlatform {
 								[&pnIndices](auto& el) -> size_t { return pnIndices ? pnIndices.get()[el - 1] - 1 : el - 1; });
 							polygon->addFace(std::begin(indices), std::end(indices));
 
-							if (face.template isOfType<typename IfcEntityTypesT::IfcIndexedPolygonalFaceWithVoids>())
-								BLUE_LOG(warning) << "Ignoring voids at " << face->getErrorLog();
+							if (face.template isOfType<typename IfcEntityTypesT::IfcIndexedPolygonalFaceWithVoids>()) {
+								auto faceWithVoids = face.template as<typename IfcEntityTypesT::IfcIndexedPolygonalFaceWithVoids>();
+
+								std::vector<std::pair<size_t, size_t>> indicesWithVoids;
+								std::vector<size_t> secondIndices(faceWithVoids->InnerCoordIndices.size());
+
+								std::transform(std::begin(faceWithVoids->InnerCoordIndices[0]), std::end(faceWithVoids->InnerCoordIndices[0]), std::begin(secondIndices),
+									[&pnIndices](auto& el) -> size_t { return pnIndices ? pnIndices.get()[el - 1] - 1 : el - 1; });
+								for (size_t i = 0; i < faceWithVoids->CoordIndex.size(); ++i)
+								{
+									indicesWithVoids[i].first = indices[i];
+								}
+
+								for (size_t i = 0; i < faceWithVoids->InnerCoordIndices.size(); ++i)
+								{
+									indicesWithVoids[i].second = secondIndices[i];
+								}
+								
+							}
+								
+
 						}
 
 						itemData->open_or_closed_polyhedrons.push_back(polygon);
--- a/Core/src/IfcGeometryConverter/FaceConverter.h
+++ b/Core/src/IfcGeometryConverter/FaceConverter.h
@@ -1246,13 +1246,12 @@ namespace OpenInfraPlatform {
 					{
 						std::shared_ptr<carve::input::PolyhedronData> polygon(new carve::input::PolyhedronData());
 						auto faceSet = tessItem.template as<typename IfcEntityTypesT::IfcPolygonalFaceSet>();
-
-						double length_factor = UnitConvert()->getLengthInMeterFactor();
-
+						
 						// obtain vertices from coordinates list and add them to the new polygon
 						for (const auto& point : faceSet->Coordinates->CoordList)
 						{
-							carve::geom::vector<3> vertex =
+							
+							carve::geom::vector<3> vertex = //placementConverter->convertIfcPoint(point);
 								carve::geom::VECTOR(point[0],
 									point[1],
 									point[2]) * UnitConvert()->getLengthInMeterFactor();
@@ -1285,14 +1284,19 @@ namespace OpenInfraPlatform {
 									[&pnIndices](auto& el) -> size_t { return pnIndices ? pnIndices.get()[el - 1] - 1 : el - 1; });
 								for (size_t i = 0; i < faceWithVoids->CoordIndex.size(); ++i)
 								{
-									indicesWithVoids[i].first = indices[i];
+									indicesWithVoids[i].first = 0;
+									indicesWithVoids[i].second = indices[i];
 								}
 
 								for (size_t i = 0; i < faceWithVoids->InnerCoordIndices.size(); ++i)
 								{
-									indicesWithVoids[i].second = secondIndices[i];
+									indicesWithVoids[i + faceWithVoids->CoordIndex.size()].first = 1;
+									indicesWithVoids[i + faceWithVoids->CoordIndex.size()].second = secondIndices[i];
 								}
 								
+								std::vector<carve::geom2d::P2> merged_path;
+								std::vector<carve::triangulate::tri_idx> triangulated;
+								std::vector<std::pair<size_t, size_t>> path_all_loops;
 							}
 								
 
--- a/Core/src/IfcGeometryConverter/FaceConverter.h
+++ b/Core/src/IfcGeometryConverter/FaceConverter.h
@@ -1289,7 +1289,7 @@ namespace OpenInfraPlatform {
 									size_t i = 0;
 									for (const auto& point : faceSet->Coordinates->CoordList)
 									{
-										if (i == outerLoop-1) {
+										if (i == outerLoop - 1) {
 											
 											carve::geom::vector<3> vertex = //placementConverter->convertIfcPoint(point);
 											carve::geom::VECTOR(point[0],
@@ -1314,7 +1314,7 @@ namespace OpenInfraPlatform {
 										double i = 0.;
 										for (const auto& point : faceSet->Coordinates->CoordList)
 										{
-											if (i == innerLoopPoint-1) {
+											if (i == innerLoopPoint- 1) {
 
 												carve::geom::vector<3> vertex = //placementConverter->convertIfcPoint(point);
 													carve::geom::VECTOR(point[0],
@@ -1333,7 +1333,15 @@ namespace OpenInfraPlatform {
 								}
 								
 								std::vector<std::vector<carve::geom::vector<2>>> loops2D;
+								std::vector<carve::geom::vector<3>> outerLoop = loops[0];
 								carve::geom::vector<3> normalOfPlane = GeomUtils::computePolygonNormal(loops[0]);
+
+								carve::math::Matrix planeMatrix;
+								GeomUtils::convertPlane2Matrix(normalOfPlane, carve::geom::VECTOR(0., 0., 0.),
+									outerLoop[2] - outerLoop[1], planeMatrix);
+
+								carve::math::Matrix inversePlaneMatrix = GeomUtils::computeInverse(planeMatrix);
+
 
 								for (int i = 0; i < loops.size(); i++) 
 								{
@@ -1341,12 +1349,8 @@ namespace OpenInfraPlatform {
 									std::vector<carve::geom::vector<2>> loop2D;
 									for (int j = 0; j < loop.size(); j++)
 									{
-										carve::geom::vector<3> point3D = loop[j];
-										carve::geom::vector<3> distance = carve::geom::VECTOR(
-											point3D.x * normalOfPlane.x,
-											point3D.y * normalOfPlane.y, 
-											point3D.z * normalOfPlane.z);
-										carve::geom::vector<3> point3Dto2D = loop[j] - distance;
+										carve::geom::vector<3> point3Dto2D = inversePlaneMatrix * loop[j];
+										
 										carve::geom::vector<2> point2D = carve::geom::VECTOR(point3Dto2D.x, point3Dto2D.y);
 										loop2D.push_back(point2D);
 									}
--- a/Core/src/IfcGeometryConverter/FaceConverter.h
+++ b/Core/src/IfcGeometryConverter/FaceConverter.h
@@ -1282,7 +1282,9 @@ namespace OpenInfraPlatform {
 
 								// 1. Build loops of actual points
 								std::vector<std::vector<carve::geom::vector<3>>> loops;
+								std::vector<carve::geom::vector<3>> outerLoop;
 								std::vector<std::vector<size_t>> indexLoops;
+								std::vector<size_t> outerIndexLoop;
 
 								for (const auto outerLoopPointIndex : faceWithVoids->CoordIndex)
 								{
@@ -1291,10 +1293,12 @@ namespace OpenInfraPlatform {
 										point[1],
 										point[2]) * UnitConvert()->getLengthInMeterFactor();
 
-									indexLoops[0].push_back(outerLoopPointIndex);
-									loops[0].push_back(vertex);
+									outerLoop.push_back(vertex);
+									outerIndexLoop.push_back(outerLoopPointIndex);
 								}
-								
+								indexLoops.push_back(outerIndexLoop);
+								loops.push_back(outerLoop);
+
 								for (const auto& innerLoop : faceWithVoids->InnerCoordIndices)
 								{
 									std::vector<carve::geom::vector<3>> loop;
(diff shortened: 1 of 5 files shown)


SamuilsRulovs:
I have moved part of the code from GeomUtils::extrude function, however I guess it is not correct right. I will be looking for a solution, how to change this function, so it would be possible to use it in both functions (in Geomutils::extrude and convertIfcIndexedFaceWithVoids).

pjanck:
Does this supersede #342 ?

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1255,14 +1255,15 @@ namespace OpenInfraPlatform {
 							carve::geom::vector<3> vertex =
 								carve::geom::VECTOR(point[0],
 									point[1],
-									point[2]) * length_factor;
+									point[2]) * UnitConvert()->getLengthInMeterFactor();
```
pjanck: There exists a function that translates list of Cartesian Points already available in Placement Converter. Please use that one.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1274,8 +1275,27 @@ namespace OpenInfraPlatform {
...
+									[&pnIndices](auto& el) -> size_t { return pnIndices ? pnIndices.get()[el - 1] - 1 : el - 1; });
+								for (size_t i = 0; i < faceWithVoids->CoordIndex.size(); ++i)
+								{
+									indicesWithVoids[i].first = indices[i];
+								}
+
+								for (size_t i = 0; i < faceWithVoids->InnerCoordIndices.size(); ++i)
+								{
+									indicesWithVoids[i].second = secondIndices[i];
+								}
```
pjanck: This doesn't seem right. What exactly are you trying to achieve here?

Core/src/IfcGeometryConverter/GeomUtils.cpp
```diff
@@ -150,10 +150,10 @@ void GeomUtils::extrude(
 	//  0-------face_loops[0]--------1
 
 	carve::geom::vector<3> normal_first_loop;
-	std::vector<std::vector<carve::geom2d::P2> >	face_loops;
-	for( std::vector<std::vector<carve::geom::vector<2> > >::const_iterator it_face_loops = face_loops_input.begin(); it_face_loops != face_loops_input.end(); ++it_face_loops )
+	std::vector<std::vector<carve::geom2d::P2>>	face_loops;
+	for( std::vector<std::vector<carve::geom::vector<2>>>::const_iterator it_face_loops = face_loops_input.begin(); it_face_loops != face_loops_input.end(); ++it_face_loops )
 	{
-		const std::vector<carve::geom::vector<2> >& loop = (*it_face_loops);
+		const std::vector<carve::geom::vector<2>>& loop = (*it_face_loops);
```
pjanck: Such for loops can be greatly simplified using the notation `for( auto element : array )` with corresponding names for element and array.

Core/src/IfcGeometryConverter/GeomUtils.cpp
```diff
@@ -432,6 +395,53 @@ void GeomUtils::extrude(
 	}
 }
 
+void GeomUtils::createVoids(std::vector<carve::geom2d::P2> merged_path,
+	std::vector<carve::triangulate::tri_idx> triangulated,
+	std::vector<std::pair<size_t, size_t>> path_all_loops,
+	std::vector<std::vector<carve::geom2d::P2>>	face_loops,
+	std::stringstream& err)
```
pjanck: Either you need to pass by reference ( `std::vector<...>& name`) or you need to return the calculated values (preferable).

Core/src/IfcGeometryConverter/GeomUtils.cpp
```diff
@@ -432,6 +395,53 @@ void GeomUtils::extrude(
...
+	merged_path.reserve(path_all_loops.size());
+	for (size_t i = 0; i < path_all_loops.size(); ++i)
+	{
+		int loop_number = path_all_loops[i].first;
+		int index_in_loop = path_all_loops[i].second;
+
+		if (loop_number >= face_loops.size())
+		{
+			err << "extrude: loop_number >= face_loops_projected.size()" << std::endl;
+			continue;
```
pjanck: This could be a throw exception

Core/src/IfcGeometryConverter/GeomUtils.cpp
```diff
@@ -215,49 +215,11 @@ void GeomUtils::extrude(
...
-		//  |   2------------------>3    |
-		//  |   ^                   |    |
-		//  |   |                   v    |
-		//  |   1, 5<---------------4    |
-		//  | /                          |
-		//  0,6------------------------->7
-		carve::triangulate::triangulate(merged_path, triangulated);
-		carve::triangulate::improve(merged_path, triangulated);
-		// triangles: (9,0,1)  (5,6,7)  (4,5,7)  (4,7,8)  (9,1,2)  (8,9,2)  (3,4,8)  (2,3,8)
+		GeomUtils::createVoids(merged_path, triangulated, path_all_loops, face_loops, err);
```
pjanck: I'd name the function `incorporateVoids` since you're not creating voids, but rather just accounting for given voids.

pjanck (changes requested):
Some preliminary feedback.

SamuilsRulovs:
> Does this supersede #342 ?

Yes, it is. As I checked #342 only adds unit test for "polygonal -face-tessellation", but it does not actually adress issues  #61 and  #317

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: The type of the indicesWithVoids  should be `std::vector<std::pair<size_t, size_t>>`. (Same as path_all_loops in GeomUtils.cpp (Line 218)). I could not transform CoordIndices and InnerCoordIndices to the vector of the type `std::vector<std::pair<size_t, size_t>>`. So I decided to create two separate vectors and put them together. These for loops are doing this.

Core/src/IfcGeometryConverter/GeomUtils.cpp
SamuilsRulovs: Resolved:  71ec6b7

Core/src/IfcGeometryConverter/GeomUtils.cpp
SamuilsRulovs: Resolved:  71ec6b7

Core/src/IfcGeometryConverter/GeomUtils.cpp
SamuilsRulovs: Resolved:  71ec6b7

Core/src/IfcGeometryConverter/GeomUtils.cpp
SamuilsRulovs: Resolved: 1805f06

pjanck (commented):
As discussed, great progress!

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1332,6 +1332,21 @@ namespace OpenInfraPlatform {
...
+
+								carve::geom::vector<3> firstPoint = firstPointloop[0];
+								carve::geom::vector<3> secondPoint = firstPointloop[1];
+								carve::geom::vector<3> thirdPoint = firstPointloop[2];
+
+								carve::geom::vector<3> firstToSecond = secondPoint - firstPoint;
+								carve::geom::vector<3> firstToThird = thirdPoint - firstPoint;
+
+								carve::geom::vector<3> normalOfPlane = carve::geom::cross(firstToSecond, firstToThird);
+								normalOfPlane.normalized();
```
pjanck: I'd introduce a function that takes a vector of points and returns the normal.

Please also have a look at the code in the new function `correctWinding` - there, plane normals are already calculated. So preferably, there would be one function that both call. (I vaguely remember that such a function should already exist in GeomUtils - did you have a look?)

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1289,7 +1289,7 @@ namespace OpenInfraPlatform {
 									size_t i = 0;
 									for (const auto& point : faceSet->Coordinates->CoordList)
 									{
-										if (i == outerLoop) {
+										if (i == outerLoop-1) {
```
pjanck: this if does not make any sense - why is it there?

pjanck (changes requested):
A suggestion and a question.

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: Because faceSet->Coordinates->CoordList is a vector of coordinates, which, as we know, goes always from 0 to n-1. 
However, indices here are from 1 to n. Thus, to select correct coordinates of the point we should either subtract 1 from indices, or add 1 to i in this if statement

Core/src/IfcGeometryConverter/FaceConverter.h
pjanck: Now it even more doesn't seem right. We need to discuss this.

Core/src/IfcGeometryConverter/FaceConverter.h
pjanck: Does it work, though? I have a feeling that this is wrong. Please also observe how the indices are handled above, with no voids in the face.

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: This code is redundant. I've forgot do delete it.

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: Resolved: 4db3b19

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: Resolved:
e5e0196

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1332,22 +1332,27 @@ namespace OpenInfraPlatform {
...
+								std::vector<std::vector<carve::geom::vector<2>>> loops2D;
+								carve::geom::vector<3> normalOfPlane = GeomUtils::computePolygonNormal(loops[0]);
 
+								for (int i = 0; i < loops.size(); i++) 
+								{
+									std::vector<carve::geom::vector<3>> loop = loops[i];
+									std::vector<carve::geom::vector<2>> loop2D;
+									for (int j = 0; j < loop.size(); j++)
+									{
+										carve::geom::vector<3> point3D = loop[j];
```
pjanck: Propose to use:

`for( const auto point3D : loop )`

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1332,22 +1332,27 @@ namespace OpenInfraPlatform {
...
-
-								carve::geom::vector<3> normalOfPlane = carve::geom::cross(firstToSecond, firstToThird);
-								normalOfPlane.normalized();
-								
+								std::vector<std::vector<carve::geom::vector<2>>> loops2D;
+								carve::geom::vector<3> normalOfPlane = GeomUtils::computePolygonNormal(loops[0]);
 
+								for (int i = 0; i < loops.size(); i++) 
+								{
+									std::vector<carve::geom::vector<3>> loop = loops[i];
```
pjanck: similarly:

`for( const auto loop : loops )`

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1332,22 +1332,27 @@ namespace OpenInfraPlatform {
...
+									std::vector<carve::geom::vector<2>> loop2D;
+									for (int j = 0; j < loop.size(); j++)
+									{
+										carve::geom::vector<3> point3D = loop[j];
+										carve::geom::vector<3> distance = carve::geom::VECTOR(
+											point3D.x * normalOfPlane.x,
+											point3D.y * normalOfPlane.y, 
+											point3D.z * normalOfPlane.z);
```
pjanck: Use dot product (should be available in carve library).

pjanck (changes requested):
Syntax suggestions.

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: dot product gives back a double, not a vector. Here I need scalar, I guess.

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: With: 
image

Without: 
image

As you can see, if I do not use -1,  a list of inner points have size of 3  (should have 4) + outer points are not on the one plane

Core/src/IfcGeometryConverter/FaceConverter.h
pjanck: Right you are, my bad! It would be better to use a matrix here in order to "project" the vector on the plane.

Core/src/IfcGeometryConverter/FaceConverter.h
pjanck: interesting ... :) well, as long as it works, I don't mind any solution.

Core/src/IfcGeometryConverter/FaceConverter.h
pjanck: Use [this help](https://stackoverflow.com/questions/9605556/how-to-project-a-point-onto-a-plane-in-3d) (great sketches!) and the already existing function in `GeomUtils`.

SamuilsRulovs: Thank you!

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: Resolved:
7a17b33

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: Resolved:
7a17b33

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: Resolved:
7a17b33

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: Unfortunately, I could not find a function, which could replace this part of code. Can you tell me please, where (which line) I can find this function?

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1282,92 +1282,113 @@ namespace OpenInfraPlatform {
...
+								}
+
+								for (const auto& triangulated : triangulatedList)
+								{
+									int a = triangulated.a;
+									int b = triangulated.b;
+									int c = triangulated.c;
+
+									polygon->addFace(indexedPath[a], indexedPath[b], indexedPath[c]);
+								}
```
pjanck: This seems OK, but I'm unsure if it really works - we need to discuss this.

Core/src/IfcGeometryConverter/GeomUtils.cpp
```diff
@@ -1027,10 +1027,11 @@ bool GeomUtils::bisectingPlane(const carve::geom::vector<3>& v1,
 
 /**********************************************************************************************/
 
-void GeomUtils::convertPlane2Matrix( const carve::geom::vector<3>& plane_normal, 
-									const carve::geom::vector<3>& plane_position, 
-		const carve::geom::vector<3>& local_z, carve::math::Matrix& resulting_matrix )
+carve::math::Matrix GeomUtils::convertPlane2Matrix( const carve::geom::vector<3>& plane_normal,
+		const carve::geom::vector<3>& plane_position, 
+		const carve::geom::vector<3>& local_z)
```
pjanck: Please add a doxycomment.

pjanck (commented):
I believe we're very close :D

Core/src/IfcGeometryConverter/GeomUtils.cpp
SamuilsRulovs: Resolved:
cdd784e

Core/src/IfcGeometryConverter/FaceConverter.h
SamuilsRulovs: Resolved:
9cc48d8

SamuilsRulovs:
I am going to move IndexedPolygonalFaceWithVoids to the separate function, but overall IndexedPolygonalFaceWithVoids is working now

jschlenger (approved):
Looks fine to me

Core/src/IfcGeometryConverter/GeomUtils.cpp
```diff
@@ -1089,6 +1089,73 @@ void  GeomUtils::applyPositionToVertex(
...
+*/
+std::vector<carve::geom::vector<2>> GeomUtils::removeEmptyCoordinate(const std::vector<carve::geom::vector<3>>& listOfVectors3D) 
+{
+	std::vector<carve::geom::vector<2>> listOfVectors2D;
+	bool notEmptyCoord = false;
+	//Check if x Coordniates are empty.
+	for (auto vector : listOfVectors3D)
+	{
+		if (! vector.x == 0.)
+		//if (!this->GeomSettings()->areEqual(vector.x, 0.))
```
SamuilsRulovs, line 1109: I could not find an option how to call function areEqual from GeometrySettings.h. Can you please give me a hint?

pjanck (approved):
Great work! Awesome!

Issue #408, opened 2021-03-15
[TODO] Performance optimation in splineUtilities (derivatives)


christophKaiser:
**Location in code**
file `Core/src/IfcGeometryConverter/SplineUtilities.h`, functions of curvature and derivative calculation - especially the function `computeCurvatureOfBSplineCurveWithKnots` and its called subfunctions.

Performance optimation could be meaningful because the calculation with 168 control points needs about 1:40 minutes (observed with the option *Stroke to Alignment* in the class `SplineInterpretation`.

**Additional context**
The function call hierarchy in the calculation of the curvature looks about like the following: (simplified)
```
computeCurvatureOfBSplineCurveWithKnots
  computePointOfBSpline
    obtainBasisFunctionBSpline
  computePointOfDerivativeOne
    obtainBasisFunctionBSpline  // first recall
    obtainBasisFunctionDerivativeOne
  computePointOfDerivativeTow
    obtainBasisFunctionBSpline  // second recall
    obtainBasisFunctionDerivativeOne  // first recall
    obtainBasisFunctionDerivativeTwo
```
By restructuring the functions, probably the repeating calls of the 'obtain Basis Function ...' could be prevented. Maybe, some other optimizations are possible, as well.

Till now, the function `computeCurvatureOfBSplineCurveWithKnots` is only used by the interpretation of B-Splines to alignment elements (see #202). It isn't used in any ifc-converters, thus this task seems to be not urgent.

PR #455, opened 2021-05-14
[REF] SplineConverter - Surfaces

### LINKED ISSUE #199: [REF] SplineConverter.h
In `IfcGeometryConverter\SplineConverter.h`, the functions of B-Spline curves were refactored in pull request #178.

Refactoring is pending for the following functions of B-Spline surfaces:
- [ ] convertIfcBSplineSurface including its specialized template function
- [ ] computeBSplineSurface

At the moment, there is no clear appointment to do that - however, October 2020 might be a good time.
Till end of September, small changes on the function `computeBSplineBasisFunctions` will be done by @christophKaiser.

christophKaiser:
Goal of this pull request is the code-reactivation and refactoring to convert entities of `IfcBSplineSurface` including its sub-types. It will fix #199.

Currently, the main functionality will be developed in `SplineConverter::convertIfcBSplineSurface` to achieve a working code. Some out-commented code fragments were found in the `FaceConverter`, and some functions of `SplineUtilities` will be used, as well. The distribution between these classes (especially `SplineConverter` and `FaceConverter`) will be probably decided later during developing.


christophKaiser:
The effort to get the control points in a `std::vector<std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcCartesianPoint>>>` seems to be quit high - at least in terms of source code (see 75811a2 or below). Did I do something to complicated at this place or missed an already exiting function to do that?
https://github.com/tumcms/Open-Infra-Platform/blob/41df2e1639b87691fe91126ebe9b86ce6435a630/Core/src/IfcGeometryConverter/SplineConverter.h#L170-L189

The function `FaceConverter::convertIfcCartesianPoint2DVector` seems suitable to convert that vector to a vector of `carve::geom::vector<3>`. This function wasn't called from any code, before. I made slight changes to get it to compile and to work (41df2e1). Probably, further refactoring is reasonable to reuse the function `CurveConverter::convertIfcCartesianPointVector` or at least `PlacementConverter::convertIfcCartesianPoint`.

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -147,12 +147,149 @@ namespace OpenInfraPlatform
 					 * \internal	The Code of the function is in the commented out part at the end of the file.
 					 */
 					void convertIfcBSplineSurface(
-						const std::shared_ptr<typename IfcEntityTypesT::IfcBoundedSurface>& splineSurface,
-						const std::vector<std::vector<carve::geom::vector<3>>>& controlPoints,
+						const EXPRESSReference<typename IfcEntityTypesT::IfcBSplineSurface>& splineSurface,
+						const carve::math::Matrix& pos, // AT THE MOMENT, NO IDEA WHAT THIS IS
```
pjanck: `pos` is the position of the geometry in the global context (is specified elsewhere in the IFC file and thus provided as a parameter to the function)  and should be accounted for when interpreting the geometries. This is a parameter in almost all functions.

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -147,12 +147,149 @@ namespace OpenInfraPlatform
...
+							const int orderV = degreeV + 1;
+							// obtain number of control points
+							//const int numControlPointsU = controlPoints.size();
+							//const int numControlPointsV = controlPoints[0].size();
+
+							// load array of control points from ifc-entity
+							// simplified, it's in the type EXPRESSContainer<EXPRESSContainer<IfcCartesianPoint>>
+							const auto controlPointList = bsplineSurfaceWithKnots->ControlPointsList;
+							// temporary control point container of type std::vector<std::vector<..>> which is needed for converter function
+							std::vector<std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcCartesianPoint>>> controlPointListVector;
```
pjanck: An `EXPRESSContainer` is a `std::vector` so there should be no need to transform here. (I may be wrong, though).

Core/src/IfcGeometryConverter/SplineConverter.h
```diff
@@ -147,12 +147,149 @@ namespace OpenInfraPlatform
...
+								}
+							}
+
+							// temporary instances to curve and face converter, potential relocation will be considered in further refactoring
+							std::shared_ptr<CurveConverterT<IfcEntityTypesT>> curveConverter =
+								std::make_shared<CurveConverterT<IfcEntityTypesT>>(CurveConverterT<IfcEntityTypesT>(GeomSettings(), UnitConvert(), placementConverter));
+							FaceConverterT<IfcEntityTypesT> faceConverter(GeomSettings(), UnitConvert(), placementConverter, curveConverter);
+
+							// obtain control points
+							const std::vector<std::vector<carve::geom::vector<3>>> controlPoints = faceConverter.convertIfcCartesianPoint2DVector(controlPointListVector);
```
pjanck: I see that the transformation from `EXPRESSContainer` to `std::vector` is necessary because of this function call. I suggest changing the function to take `EXPRESSContainer`.

pjanck (changes requested):
Good progress - some small suggestions from me.

pjanck:
> Did I do something to complicated at this place or missed an already exiting function to do that?

Nope, your approach seems OK. This area of code did not receive any love before, therefore it seems so strange. But you're on the right track!

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: By puting `bsplineSurfaceWithKnots->ControlPointsList` directly into a `std::vector<std::vector<...>>`, the compiler complained " 'initializing': cannot convert from 'const OpenInfraPlatform::EarlyBinding::EXPRESSContainer<T,0,18446744073709551615>' to 'std::vector<std::vector<ValueType,std::allocator<_Ty>>,std::allocator<std::vector<_Ty,std::allocator<_Ty>>>>' "

However, at least this specific case is resolved by the conversation next below (about changing the function parameter type).

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: Done in commit 9b4c92d.

christophKaiser:
Added a further suggestion of refactoring the function `FaceConverter::convertIfcCartesianPoint2DVector` in commit c5322c6: The function design is adapted from `CurveConverter::convertIfcCartesianPointVector` which converts a list by spliting into single points. The refactored function works similar by spliting an array into several lists.
The treatment of `lengthFactor` and number of dimension are handled by the called function `PlacementConverter::convertIfcCartesianPoint`.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1032,7 +1032,9 @@ namespace OpenInfraPlatform {
 				*/
 
 				std::vector<std::vector<carve::geom::vector<3>>> convertIfcCartesianPoint2DVector(
-					const std::vector<std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcCartesianPoint>>>& points2D)  const throw(...)
+					const OpenInfraPlatform::EarlyBinding::EXPRESSContainer<
+					OpenInfraPlatform::EarlyBinding::EXPRESSContainer<
+					EXPRESSReference<typename IfcEntityTypesT::IfcCartesianPoint>, 0, -1>, 0, -1>& points2D)  const throw(...)
```
christophKaiser: Currently, it works. However, the long typename
`OpenInfraPlatform::EarlyBinding::EXPRESSContainer<OpenInfraPlatform::EarlyBinding::EXPRESSContainer<EXPRESSRefrence<...`
looks a bit odd.
A simple reduction to `EXPRESSContainer<EXPRESSContainer<EXPRESSReference<...` does not compile because of "missing type specifier".
Does someone has a suggestion how to improve this?

pjanck: All `EXPRESS...` classes are added to the `oip` namespace, so sth like `oip::EXPRESSContainer` should work.

christophKaiser: Currently, `oip::EXPRESSContainer` is not available in the `oip` namespace:
210515 oip_EXPRESS__

I found three possible solutions:
- Using of the existing type alias `using LIST` from [EXPRESSContainer.h](https://github.com/tumcms/Open-Infra-Platform/blob/76cd098fdc663ddf1983f214e559f1b92d91a0f0/EarlyBinding/src/EXPRESS/EXPRESSContainer.h#L233). The variable declaration in the `FaceConverter` would look like
`oip::LIST<0, -1, oip::LIST<0, -1, EXPRESSReference<...>>>`
This solution is not very nice and intuitive because the `EXPRESSContainer` is hidden behind the alias `LIST`.
- Including `EXPRESSContainer` to the namespace `oip` by adding line 245 in EXPRESSContainer.h:
210516 oip_EXPRESSContainer into namespace
The variable declaration in the `FaceConverter` would look like
`oip::EXPRESSContainer<oip::EXPRESSContainer<EXPRESSReference<...>, 0, -1>, 0, -1>`
This is similar to the way how all other `EXPRESS...` classes are handled.
- Adding of a new type alias in line 240 of EXPRESSContainer.h
210516 oip_EXPRESSContainer using alias
The variable declaration in the `FaceConverter` would look like
`EXPRESSContainer<EXPRESSContainer<EXPRESSReference<...>>>`
This is the way how the `EXPRESSReference` type is used almost everywhere. Probably, it could be the simplest and most intuitive declaration for instances of `EXPRESSContainer`.

I want to ask for some advice to this suggestions because of the possible big side effects of changes in the `EXPRESS...` classes (which was generally mentioned in a previous dev-meeting).

pjanck: For the time being, use option 3 but add the typename definition in the FaceConverter.h file (NOT EXPRESSContainer). We need to discuss this further, but this should allow you to continue programing.

Also: use (similarly to the definition of the `ControlPointsList`) 0 and LIST_MAXSIZE as the integer template parameters.

jschlenger (commented):
Wasn't able to detect any errors other than those mentioned by pjanck

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 48612b2

christophKaiser:
Commit 0b515e0 implements the setup in the `FaceConverter` to represent the inheritance structure of `IfcBSplineSurface` and `IfcBSplineSurfaceWithKnots`. The parameter type and return type of `SplineConverter::convertIfcBSplineSurface` is adjusted, as well. Commit 044a541 adjusts the function name to a more suitable one; additionally this should prevent confusion with the function `FaceConverter::convertIfcBSplineSurface`.
Commit 48d2604 moves the obtainment of the control points into the `FaceConverter`, which makes obsolete the temporary instance of `FaceConverter` inside the `SplineConverter`. The control points are loaded based on an `IfcBSplineSurface` entity.

The commits 31c3e93 and 51241cd changed the existing function of `loadKnotArray` to be reused for the surface. The call structure of the functions looks like the following schema:
210523 obtainKnotArray schema

Commit 4bfa5a6 prepares the structure to distinguish between `IfcBSplineSurfaceWithKnots` and `IfcRationalBSplineSurfaceWithKnots`; its content is in process.

christophKaiser:
**Most reason changes:**
(not all commits will be named individually)

Commits 4ce7377 and cbfd66a changed the loading process of weights for surfaces: Now, the existing function `loadWeightsData` receives a vector of `IfcReals` instead of one `EXPRESSReference` entity, thus it can be reused by loading the array of weights in case of a surface.

Commit 299dc6d splits the function `obtainProperties` into multiple functions, the new function `obtainNumCurvePoints` is called twice by the surface-converter.
- [x] At this point, the "TEMPORARY default number of curve points" might need further discussion. Should it stay that way, at least for now? Or from where / how the actual number should be determined?

Commits 957bd7a and 4f1a128 implement the functions for the calculation of B-Spline-Surfaces and rational-B-Spline-Surfaces. Both sets of functions are quite similar, but with difference of having a `weights` array, and different internal formulars (steps 3a and 3b). This segregation into two sets of functions was with the attempt of _only one task per function_ (like it was the case in curve calculations, as well).
- [x] However, a combination into one function with differentiation between the two cases should be discussed, because of the large similar parts of code. Maybe, it could look like the legacy code of `SplineConverter` which is removed in commit 61195a1.
The `weights` array will be still a input parameter, but pre-set with an empty- / nonsense- default value in the function declaration - in case we consider this combination, further research and test might be required on my side.

I took a view in more detail about the meaning of the parameter `pos`. Commit c627ad8 applies it to the curve points, this resolves an open conversation above. Thanks @pjanck for the given search keywords in the private message!

Commit a373272 implements a suggestion from my side how to distribute the tasks to the three classes:
- `FaceConverter`: Identification of an `IfcBSplineSurfaceWithKnots`, loading of control points form ifc-entity (because a called sub-function is already there), construction of face point relationships by `PolylineSetData`
- `SplineConverter`: Loading and preparation of all properties form the ifc-entity, apply of the position transformation
- `SplineUtilities`: Mathematical B-Spline calculations (independent from ifc-entities / `EXPRESSReference`)

EDIT 11th June: Updated checkbox.
EDIT 15th June: Updated checkbox.

Core/src/IfcGeometryConverter/SplineConverter.h
christophKaiser: I took a more detaild look into it. Thanks for the research keywords in the private message!
Done in commit c627ad8.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -282,46 +282,113 @@ namespace OpenInfraPlatform {
...
-					*/
+						// declaration of a list, for each curve point (=face point) its x, y, z coordinate will be stored as string with an id
+						std::unordered_map<std::string, int> vertexMap;
+						vertexMap.reserve(numCurvePointsU * numCurvePointsV);
+
+						// loop over all curve points
+						for (int v = 0; v < numCurvePointsV - 1; ++v)
+						{
+							for (int u = 0; u < numCurvePointsU - 1; ++u)
+							{
```
jschlenger: It's recommended to used size_t instead of int for variables that are used as an index

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -282,46 +282,113 @@ namespace OpenInfraPlatform {
...
+								// A--->B    0-->u
+								//
+								// there is no closing line from D to 
+
+								// start a new poly line
+								polylineData->beginPolyline();
+
+								// loop over the 4 conter points
+								for (auto k = 0; k < 4; ++k)
+								{
```
jschlenger: better use size_t (see comment before)

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1020,56 +1087,29 @@ namespace OpenInfraPlatform {
...
+				The alias is defined for use in the function parameter of \c convertIfcCartesianPoint2DVector.
 				*/
+				template <typename T> using EXPRESSContainer = OpenInfraPlatform::EarlyBinding::EXPRESSContainer<T, 0, LIST_MAXSIZE>;
 
+				/*! \brief Converts a two-dimensional array of \c IfcCartesianPoint-s to a array of \c carve points.
+				\param	points2D	The two-dimensional array of \c IfcCartesianPoint-s to be converted.
+									In the \c ifc -documentation, it's called i. e. 'a list of lists of control points'.
+				\returns			The two-dimensional array of \c carve points.
+				*/
 				std::vector<std::vector<carve::geom::vector<3>>> convertIfcCartesianPoint2DVector(
```
jschlenger: maybe naming the function "convertIfcCartesianPoint2Vector2D" could be a little bit clearer

jschlenger (commented):
Sorry for the late review.

Looks like some good progress!
Just a few minor comments.

I think for now a default number of curvePoints is absolutely fine. From ifcClothoid I know that the number of points is dependent on the curvature but I guess this would be more complicated for your case. Maybe the number of points could be dependent on the total size of the bspline/ bsplinesurface. But I would stick to a default number for now and think about this later on.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -282,52 +282,113 @@ namespace OpenInfraPlatform {
 						\note	The \c IfcBSplineSurface is a subtype of \c IfcBoundedSurface.
 					*/
 
-				std::shared_ptr<carve::input::PolylineSetData> convertIfcBSplineSurface(const EXPRESSReference<typename IfcEntityTypesT::IfcBSplineSurface>& surface,
+				std::shared_ptr<carve::input::PolylineSetData> convertIfcBSplineSurface(
+					const EXPRESSReference<typename IfcEntityTypesT::IfcBSplineSurface>& surface,
 					const carve::math::Matrix& pos) const throw(...)
```
pjanck: As discussed during our last meeting, make this a void and take in another parameter `itemData` - similarly to other convert functions.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -282,52 +282,113 @@ namespace OpenInfraPlatform {
...
 					if (surface.expired()) {
 						throw oip::ReferenceExpiredException(surface);
 					}
 
-					// temporary call of SplineConverterT::convertIfcBSplineSurface(..)
-					SplineConverterT<IfcEntityTypesT> splineConverter(GeomSettings(), UnitConvert(), placementConverter);
-					std::shared_ptr<carve::input::PolylineSetData> polylineData;
-					splineConverter.convertIfcBSplineSurface(surface, pos, polylineData);
+					// obtain control points, attribute 3 of IfcBSplineSurface
+					const std::vector<std::vector<carve::geom::vector<3>>> controlPoints = convertIfcCartesianPoint2DVector(surface->ControlPointsList);
```
pjanck: As discussed, a better name would be `convertIfcCartesianPointVectorVector`

pjanck (changes requested):
Regarding your questions:

1. please implement the requested changes as discussed during our live meeting
2. if the code is quite similar for both types of B-Splines, you can of course handle them in an individual function - DRY principle wins over having a separate function for each IFC entity
3. the temporary number of tessellation points - just leave it at some value (as @jschlenger said) - we need to readdress this, though
4. I agree with the split of the functions to the corresponding files

Additionally:

- [ ] add unit tests from your testing files
- [ ] check doxy-comments of the functions (you have already provided a lot of them, however, I'm requesting changes to their signatures, so it's best to double check)

I imagine that the work for `IfcAdvancedBrep` could be part of a new PR?

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: According to the individual meeting and pjanck's suggestion, it's changed to `convertIfcCartesianPointVectorVector`.
Done in commit 36d839f.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 36d839f.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 50d1da2.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 50d1da2.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 3ae78cd.

christophKaiser:
As I distributed the changes to several commits (to give them some kind of change structure) and implemented the additional parts, I was interrupted by an unexpected live lecture (including a time-consuming preparation for it). That's why four commits now popped up before your comment, @pjanck.

In commit 840bd7d, the construction of polyhedrons is adapted to the function `convertIfcPlane` in a similar way like in `convertIfcBSplineSurface` (commit 0787fe7). At least for the example file about the `IfcAdvancedBrep` (which will be a new PR),  the `IfcPlane` will be needed as a surface. I put this change in this PR, because here the restructuring to `polyhedrons` was done.

About the doxy-comments, I'm not sure whether you wanted to (implicit) refer to something specific. In commit 7286264 are some changes in doxy-comments.

The unit tests will be pushed soon.

pjanck (commented):
Great changes, commenting so that I see what I've reviewed so far.

christophKaiser:
Commit 6253f17 adds a suggestion about reducing repetitive code. However, it breaks in some way the rule of 'only one task per function' because two types of B-Splines are now handled in one function.
If this doesn't fit in this case, it's no problem to revert this commit - at least, it was a nice personal experience of default parameters in practise.

christophKaiser:
Unit Tests are added by commits 897d9c9, 1d21e67, and 569f9a8.
The markdown table of supported ifc representations is updated in commit 1200977.

From my perspective, the refactoring of the `SplineConverter` seems to be finished with these commits.
However, a performance optimisation lefts (see already existing issue #408); this become noticeable especially during the computation of B-Spline-Surfaces (here it's just a lag of a few seconds).

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -1125,7 +1125,7 @@ namespace OpenInfraPlatform {
 					loop.reserve(points.size());
 
 					// convert each row of points, save in target vector
-					for (auto itPoints : points)
+					for (auto& itPoints : points)
```
pjanck: It is even better to use `const auto&` in such cases, where the iterator itself is not changed in the loop.

pjanck (approved):
Looks great! Love the examples!

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 0fad7a4.

jschlenger (approved):

PR #462, opened 2021-06-19
IfcAdvancedBrep for example file cube-advanced-brep

### LINKED ISSUE #58: Example file "basin-advanced-brep" is not working
The file is [here](../blob/development/testdata/IFC4x1ExampleFiles/official/E3_Advanced_geometric_shape/basin-advanced-brep.ifc).  Current version crashes.

### LINKED ISSUE #60: Example file "cube-advanced-brep" is not working
The file is [here](../blob/development/testdata/IFC4x1ExampleFiles/official/E3_Advanced_geometric_shape/cube-advanced-brep.ifc).  Current version crashes.

christophKaiser:
In this pull request, `IfcAdvancedBrep` will be implemented to run the example file cube-advanced-brep.ifc.
This includes the implementation and adjustment of a few other functions, too. A rough overview of the involved functions was given in [a comment](https://github.com/tumcms/Open-Infra-Platform/issues/60#issuecomment-847929166) in the corresponding issue #60.

Finally, this PR will close #60.
~~Probably, issue #58 could be closed, as well. However, this will be investigated at the right time in future. [currently, the non-linked issue is intended]~~

UPDATE 21.02.2022: This PR will close #58, as well.


christophKaiser:
While I have analysed the function `convertIfcFaceList` in commit 346e611 and 032606c, I came over a 'TODO Stefan':
https://github.com/tumcms/Open-Infra-Platform/blob/1f1fc35d5c265a7c28675fe82872cd1a7595dfac/Core/src/IfcGeometryConverter/FaceConverter.h#L750-L766
(Note: The cited code is without the new changes.)

It was added in commit 2aec12f last summer. However, I wasn't able to find more information about that.
Currently, the function does its purposed task.

@pjanck Should I leave that comment there for the moment, or can you give me more information about it?

My next task in this PR will be an analyse of the function `convertIfcFace`. This will lead in a (probably) small refactoring of this function and the implementation of its subtype `IfcFaceSurface`, which is the supertype of `IfcAdvancedFace` (which I will need at the end).

pjanck:
Good analysis of the task!

1. About the `TODO Stefan`: no idea what I meant there. If the function does its task, you can safely delete it. 
2. You can change the `for` loop to `for( const auto& face : faces )` - perhaps that was the only purpose of the TODO?
3. `convertIfcFace` - it seems that the function returns `bool` - this can be changed to be more in parallel with other functions. I trust you here to judge by yourself what to do. Perhaps it is needed, that a `bool` is returned. In case of doubt, ask.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -748,10 +747,8 @@ namespace OpenInfraPlatform {
 					std::map<std::string, uint32_t> polygonIndices;
 
 					// Loop through all faces
-					for (auto it = faces.cbegin(); it != faces.cend(); ++it) 
+					for (const auto& face : faces) 
 					{ //TODO Stefan
-						EXPRESSReference<typename IfcEntityTypesT::IfcFace> face = (*it);
-
```
pjanck: I commented before looking at the code - you've already done task 2. :)

pjanck (commented):
Good start!

christophKaiser:
[Refactoring of convertIfcFace is still in process.]

Like promised in commit 3ae4b05 “Fixed semi-bug of double triangulation of triangular ifcFaceBound”, this comment gives more details about the fixed issue – especially screenshots of an example.

Before this commit, faces with an exactly triangular bound loop (`ifcFaceBound`) were added twice to the variable `polygon` (of type `PolyhedronData`). First, the triangle face was added in the if-block `if (loopVertices3D.size() == 3)` (in `convertIfcFace`). The second adding occurred inside of the function `triangulateFace` which is called in `convertIfcFace`. This became implicitly visible by the variable `polygon.faceCount`, respectively `polygon.faceIndices`.
Moreover, the first adding has not gave respect to holes inside the outer triangle bound, respectively they were covered by the first added triangle face as shown in the following screenshot:
screenshot fix double triangles 1 before

After this fix, the code decides between the case of only one triangular bound loop without an inner bound (= hole), and the case of an arbitrary number of vertices and / or holes inside the face. Now, the hole in my example is visible:
screenshot fix double triangles 2 after

The relevant ifc-data of my example:
```
#38= IFCFACE((#202,#203));

#202= IFCFACEOUTERBOUND(#366,.T.);
#203= IFCFACEBOUND(#367,.T.);

#366= IFCPOLYLOOP((#527,#526,#606));
#367= IFCPOLYLOOP((#528,#608,#529));

#526= IFCCARTESIANPOINT((0.,0.,0.));
#527= IFCCARTESIANPOINT((1000.,0.,0.));
#606= IFCCARTESIANPOINT((0.,1000.,0.));

#528= IFCCARTESIANPOINT((250.,250.,0.));
#529= IFCCARTESIANPOINT((500.,250.,0.));
#608= IFCCARTESIANPOINT((250.,500.,0.));
```

jschlenger (commented):
Good start. Excellent description of what needs to be supported in order to be able to display the example file.

christophKaiser:
The refactoring of the function `convertIfcFace` is done - at least for the moment. (Of course, the doxygen-comments will be updated soon.)
Maybe, its sub-function `addArbitraryFaceToPolyhedronData` could still need some effort.

Next, I will implement the functions for `IfcFaceSurface` and `IfcAdvancedFace` (both are sub-types of `IfcFace`). During or after this, I can probably decide whether [parts of] `addArbitraryFaceToPolyhedronData` could be reused there.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -726,7 +726,8 @@ namespace OpenInfraPlatform {
 
 				//------------------------------------------------------------------------------------------------------------
 				// FaceConverter functions:
-				// convertIfcFaceList, convertIfcFace, convert3DPointsTo2D, triangulateFace, convertIfcCartesianPointVectorVector
+				// convertIfcFaceList, convertIfcFace, convertIfcFaceBoundList, covnertIfcFaceBound, addTriangleToPolyhedronData, 
+				// addArbitraryFaceToPolyhedronData, convert3DPointsTo2D, triangulateFace, convertIfcCartesianPointVectorVector
 				//------------------------------------------------------------------------------------------------------------
```
pjanck: This can safely be deleted, it adds no value.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -748,15 +749,13 @@ namespace OpenInfraPlatform {
 
 					// Loop through all faces
 					for (const auto& face : faces) 
-					{ //TODO Stefan
+					{
 						if (face.expired()) {
 							throw oip::ReferenceExpiredException(face);
 						}
```
pjanck: This should be checked as the first thing in the `convertIfcFace` function, no need to do it here.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -768,132 +767,165 @@ namespace OpenInfraPlatform {
...
 
-					// Loop through all boundary definitions
-					int boundID = -1;
+							// according to ifc-documentation, only one boundary (= loop) can be outer boundary:
+							// "One loop is optionally distinguished as the outer loop of the face."
+							// (https://standards.buildingsmart.org/IFC/DEV/IFC4_3/RC3/HTML/link/ifcface.htm)
+							// thus, break the for-loop after THE outer boundary was found
+							break;
+						}
+					}
```
pjanck: I propose to make this a distinct function.

Takes in `std::vector<EXPRESSReference<... IfcFaceBound>>& faceBound` and puts the outer bound at index 0.

This way, the input to `convertIfcFaceBoundList`  can be made const reference (`const ...&`).

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -768,132 +767,165 @@ namespace OpenInfraPlatform {
...
-					modBounds.reserve(2);
-					bool faceLoopReversed = false;
+						addArbitraryFaceToPolyhedronData(face, faceBoundLoops, polygon, polygonIndices);
+					}
+				}
+				
+				void convertIfcFaceBoundList(
+					std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>> ifcFaceBounds,
+					const carve::math::Matrix& pos,
+					std::vector<std::vector<carve::geom::vector<3>>>& faceBoundLoops) const noexcept(true)
```
pjanck: Perhaps the `faceBoundLoops` can be made a return parameter?

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -768,132 +767,165 @@ namespace OpenInfraPlatform {
...
+					{
+						// get vertices of one bound loop into faceBoundLoops[i]
+						convertIfcFaceBound(ifcFaceBounds[i], pos, faceBoundLoops[i]);
+					}
+				}
+
+				void convertIfcFaceBound(
+					const EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>& bound,
+					const carve::math::Matrix& pos,
+					std::vector<carve::geom::vector<3>>& loopVertices3D) const noexcept(false)
```
pjanck: Similarly to above, `loopVertices3D` can be made return parameter?

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -768,132 +767,165 @@ namespace OpenInfraPlatform {
...
+				{
+					//	IfcLoop   has subtypes   IfcEdgeLoop, IfcPolyLoop, IfcVertexLoop
+					const EXPRESSReference<typename IfcEntityTypesT::IfcLoop>& loop = bound->Bound;
+					bool polyOrientation = bound->Orientation;
+
+					if (!loop) {
```
pjanck: test, if reference expired

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -768,132 +767,165 @@ namespace OpenInfraPlatform {
...
+						vertex = pos * vertex;
 					}
-					modBounds.shrink_to_fit();
 
-					for (const auto& bound : modBounds) 
+					if (loopVertices3D.size() < 3)
 					{
-						boundID++;
+						throw oip::InconsistentGeometryException(loop, " Number of vertices < 3");
+					}
```
pjanck: You can test this before the for-loop applying the position.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -768,132 +767,165 @@ namespace OpenInfraPlatform {
...
+					if (!polyOrientation)
+					{
+						std::reverse(loopVertices3D.begin(), loopVertices3D.end());
+					}
 
-						//	IfcLoop <- IfcEdgeLoop, IfcPolyLoop, IfcVertexLoop
-						const EXPRESSReference<typename IfcEntityTypesT::IfcLoop>& loop = bound->Bound;
-						bool polyOrientation = bound->Orientation;
+					return;
+				}
```
pjanck: No need for a `return` statement at the end of a `void` function (it happens nevertheless).

pjanck (commented):
Some suggestions for improvement. Otherwise, seems fine - good progress!

jschlenger (commented):
Not able to spot any additional mistakes

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 93d2916.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit b6148c3.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit b4c24cb.
However, the input of the new distinct function `swapOuterBoundaryToFront` can't be a reference to a constant vector because the function `std::rotate(...)` performs changes on this vector.

An alternative approach to avoid the operation on the vector (and prevent the copy with the swap of elements) could be to only indicate the index of the outer boundary. In the function `convertIfcFaceBoundList`, the outer boundary would be added as first element to the target `faceBoundLoops`, and then the for-loop would add all other loops (with skipping the outer boundary).
At least in terms of comprehensible code, probably this would be more complex.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit d9303cc.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit ff1b65e.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 785881d.
I added the expired-test of the `EXPRESSReference<... IfcFaceBound> bound` because it wasn't done before.
As well, I adjusted and kept the expired-test of the `EXPRESSReference<... IfcLoop> loop`, which is done immediately after it's definition. At this point, I'm not sure about it - I would expect this test at the beginning of the function `CurveConverter::convertIfcLoop`, however I can't spot it there.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 72e884d.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Indirectly done in commit ff1b65e, becouse the function type was changed from `void` to a vector.
Done for other return statements in commit 3f4c077.

Core/src/IfcGeometryConverter/FaceConverter.h
pjanck: not const is perfect, simple code is preferred :)

Core/src/IfcGeometryConverter/FaceConverter.h
pjanck: if the test is not at the beginning of the function, please put it there

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -801,59 +790,73 @@ namespace OpenInfraPlatform {
...
+					{
+						// get vertices of one bound loop into faceBoundLoops[i]
+						faceBoundLoops[i] = convertIfcFaceBound(faceBounds[i], pos);
+					}
+
+					return faceBoundLoops;
+				}
+
+				std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>> swapOuterBoundaryToFront(
+					std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>> ifcFaceBounds) const throw(...)
```
pjanck: if you're doing swaps in place, no need to return, use reference `&`
```suggestion
				void swapOuterBoundaryToFront(
					std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>>& ifcFaceBounds) const throw(...)

```

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -801,59 +790,73 @@ namespace OpenInfraPlatform {
...
+				{
+					const std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>> faceBounds = swapOuterBoundaryToFront(ifcFaceBounds);
+
+					std::vector<std::vector<carve::geom::vector<3>>> faceBoundLoops;
+					faceBoundLoops.resize(faceBounds.size());
+					for (size_t i = 0; i < faceBounds.size(); i++)
+					{
+						// get vertices of one bound loop into faceBoundLoops[i]
+						faceBoundLoops[i] = convertIfcFaceBound(faceBounds[i], pos);
+					}
```
pjanck: you could use `std::transform` here (but not necessary to change; you can learn, however)

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -801,59 +790,73 @@ namespace OpenInfraPlatform {
...
+
 					//	IfcLoop   has subtypes   IfcEdgeLoop, IfcPolyLoop, IfcVertexLoop
 					const EXPRESSReference<typename IfcEntityTypesT::IfcLoop>& loop = bound->Bound;
 					bool polyOrientation = bound->Orientation;
 
-					if (!loop) {
-						throw oip::InconsistentModellingException(bound, " No valid loop");
+					if (loop.expired()) {
+						throw oip::ReferenceExpiredException(loop);
 					}
```
pjanck: check in `convertIfcLoop`

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -801,59 +790,73 @@ namespace OpenInfraPlatform {
...
 
+					// declare target variable
+					std::vector<carve::geom::vector<3>> loopVertices3D;
+
 					// Collect all vertices of the current loop
 					curveConverter->convertIfcLoop(loop, loopVertices3D);
```
pjanck: perhaps change the return value of the function?

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -801,59 +790,73 @@ namespace OpenInfraPlatform {
...
-					std::vector<carve::geom::vector<3>>& loopVertices3D) const noexcept(false)
+					const carve::math::Matrix& pos) const noexcept(false)
 				{
+					if (bound.expired()) {
+						throw oip::ReferenceExpiredException(bound);
+					}
+
 					//	IfcLoop   has subtypes   IfcEdgeLoop, IfcPolyLoop, IfcVertexLoop
 					const EXPRESSReference<typename IfcEntityTypesT::IfcLoop>& loop = bound->Bound;
 					bool polyOrientation = bound->Orientation;
```
pjanck: do we react to this boolean?

pjanck (changes requested):
small suggestions

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 8cc1112.
The function `convertIfcEdgeLoop` will receive some futher attention within this pull request.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 5d90a20

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: A few lines later, it's used to flip the orientation / sense of the boundary.
https://github.com/tumcms/Open-Infra-Platform/blob/2715f2c87179552ecdd25ba656edf031af0fa9cb/Core/src/IfcGeometryConverter/FaceConverter.h#L849-L853
But the variable is only used once, thus commit 3aab188 would replace the variable with its actual information source.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit cc2b38a.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 7096ed8.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -790,43 +790,60 @@ namespace OpenInfraPlatform {
...
+				 * The first loop in the return list is converted from the \c IfcFaceOuterBound entity.
+				 * Only one boundary should be of this type.
+				 *
+				 * \param[in]	ifcFaceBounds	List of \c IfcFaceBound entities to be converted.
+				 * \param[in]	pos				The relative location of the origin of the representation's coordinate system within the geometric context.
+				 * \return		List of boundary loops.
+				 */
 				std::vector<std::vector<carve::geom::vector<3>>> convertIfcFaceBoundList(
-					const std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>>& ifcFaceBounds,
+					std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>> ifcFaceBounds,
```
pjanck: I'd leave the reference `&` - this way the vector is not unnecessarily copied...

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -790,43 +790,60 @@ namespace OpenInfraPlatform {
...
+					swapOuterBoundaryToFront(ifcFaceBounds);
 
 					std::vector<std::vector<carve::geom::vector<3>>> faceBoundLoops;
-					faceBoundLoops.resize(faceBounds.size());
-					for (size_t i = 0; i < faceBounds.size(); i++)
-					{
-						// get vertices of one bound loop into faceBoundLoops[i]
-						faceBoundLoops[i] = convertIfcFaceBound(faceBounds[i], pos);
-					}
+					faceBoundLoops.resize(ifcFaceBounds.size());
```
pjanck: This can done in a single line, to reserve the amount of space in a vector, I think it is this syntax:

`std::vector<std::vector<carve::geom::vector<3>>> faceBoundLoops(ifcFaceBounds.size());`

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -790,43 +790,60 @@ namespace OpenInfraPlatform {
...
 
-				std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>> swapOuterBoundaryToFront(
-					std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>> ifcFaceBounds) const throw(...)
+				/*! \brief  Swaps the \c IfcFaceOuterBound entity to the first position in the list of \c IfcFaceBound entities.
+				 * Only one boundary loop can be an \c IfcFaceOuterBound entity. But in general, it's optional. 
+				 * \param[in,out]	ifcFaceBounds	List of \c IfcFaceBound entities in which the \c IfcFaceOuterBound entity has to be swaped to the front.
+				 */
+				void swapOuterBoundaryToFront(
+					std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>>& ifcFaceBounds
+				) const throw(...)
```
pjanck: suggestion:

throw if there are more than 1 outer boundaries - you can use `std::count_if`

pjanck (changes requested):
Some suggestions

christophKaiser:
Currently, the function `convertIfcFaceBound` and its superfunction `convertIfcFaceBoundList` are implemented in the class `FaceConverter` and are called only by the function `convertIfcFace`.
However, the entity `IfcFaceBound` is a subtype of `IfcTopologicalRepresentationItem` (according to the [documentation](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_1/FINAL/HTML/link/ifcfacebound.htm)). Thus, should they be implemented in the class `ReprestationConverter` (shift the existing code into that place) and be called by the function `convertIfcTopologicalRepresentationItem` as well?

pjanck:
- no need to move the code, it is fine where it is (the functions are converting faces, right?)
- call appropriately from the `convertIfcTopologicalRepresentationItem` function - representationConverter has a faceConverter as a member (see e.g. the `convertIfcGeometricRepresentationItem`)

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: With the reference `&`, the variable `ifcFaceBounds` becoms a const vector, because the source `face->Bounds` in `convertIfcFace` (line 775) is a const variable. (`const EXPRESSReference<typename IfcEntityTypesT::IfcFace>& face`, line 763)
At least, the function `swapOuterBoundaryToFront` (called in line 805) needs a non-const vector in order to perform the swap of elements. Probably, we need the copy at some point to come from const to non-const? (see message of commit cc2b38a)
By just adding the `&` in line 802 (where you commented), the compiler complains "cannot convert argument 1 from 'const (...)' to 'std::vector<(...)> &' "

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Thanks for this hint! I had forgot about that.
Done in commit 3f00fd6.

christophKaiser:
> 
> 
>     * call appropriately from the `convertIfcTopologicalRepresentationItem` function - representationConverter has a faceConverter as a member (see e.g. the `convertIfcGeometricRepresentationItem`)

Done in commit 5c2770c (`IfcFaceBound`).
I have observed the same issue with `IfcFace` and `IfcLoop`. The calls are added similar in commits 54e028b and 01a3edc.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 5d90a20.

Core/src/IfcGeometryConverter/FaceConverter.h
pjanck: Right, then it needs copying, I see. I withdraw my comment.

Core/src/IfcGeometryConverter/RepresentationConverter.h
```diff
@@ -510,6 +510,59 @@ namespace OpenInfraPlatform {
...
+						const std::vector<carve::geom::vector<3>> vertices = faceConverter->convertIfcFaceBound(
+							topo_item.as<typename IfcEntityTypesT::IfcFaceBound>(), objectPlacement);
+
+						// Carve polyline of the converted boundary
+						std::shared_ptr<carve::input::PolylineSetData> polyline_data(new carve::input::PolylineSetData());
+						// start a new polyline
+						polyline_data->beginPolyline();
+
+						// add all vertices to polyline
+						for (auto v : vertices) {
```
pjanck:
```suggestion
						for (const auto& v : vertices) {
```

I'm not sure this would work, please try.

Core/src/IfcGeometryConverter/RepresentationConverter.h
```diff
@@ -510,6 +510,59 @@ namespace OpenInfraPlatform {
...
+						// get vertices of the loop
+						const std::vector<carve::geom::vector<3>> vertices = curveConverter->convertIfcLoop(topo_item.as<typename IfcEntityTypesT::IfcLoop>());
+
+						// Carve polyline of the converted loop
+						std::shared_ptr<carve::input::PolylineSetData> polyline_data(new carve::input::PolylineSetData());
+						// start a new polyline
+						polyline_data->beginPolyline();
+
+						// add all vertices to polyline
+						for (auto v : vertices) {
```
pjanck: similarly

pjanck (commented):
Good improvements!

pjanck:
> I have observed the same issue with IfcFace and IfcLoop. 

Very good! If you wish, you can look into `IfcVertex` as well. Should be easy to put together and some functionality should be already present in the converters.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: `throw` is introduced with commit a9990e7.
I'm not sure about the throw message: Usually, the custom `throw`s expect a single `EXPRESSReference` entity (if I understood it right). However, at this point the function has a bunch (vector) of entities which causes the throw. Just to give the user a hint about from where the error comes, I put the id of at least one entity into the string-message overload.
When trace back to a single source entity, the `IfcFace` in the function `convertIfcFace` would be the appropriate one - but currently, the function doesn't have access to it. To add it as a parameter just for the throw feels like not the best choice.

Core/src/IfcGeometryConverter/RepresentationConverter.h
christophKaiser: If work == compiles: Done in commit 363c1cf.
However, I wasn't able to run this code block with an ifc file. Probably, I will create one for an unit test of `IfcTopologicalRepresentationItem`'s a bit later (especially to test the throw statements inside the called functions and the issue of an `IfcFace` with hole which I observed [above](https://github.com/tumcms/Open-Infra-Platform/pull/462#issuecomment-864908637)).

Core/src/IfcGeometryConverter/RepresentationConverter.h
christophKaiser: Done in commit 363c1cf.

christophKaiser:
> Very good! If you wish, you can look into `IfcVertex` as well. Should be easy to put together and some functionality should be already present in the converters.

`IfcVertex` with `IfcVertexPoint` are already on my [agenda](https://github.com/tumcms/Open-Infra-Platform/issues/60#issuecomment-847929166) - they are used by the example file cube-advanced-brep. I will come to them after some effort into `IfcLoop` (with subtypes).

christophKaiser:
The current pushes implement the functions to convert an `IfcAdvancedFace`. Surprisingly, the visual result does not look wrong:

210924 MidtermResult cube
210924 MidtermResult basin

I'm surprised because there is a bunch of functions and ifc-entities on my agenda, where I want to take a detailed look into - for example `convertIfcLoop`. Maybe, they actually do the right things, or it's just luck to get right looking results. Furthermore, in those subfunctions a few ToDo's have been there before the start of this pull request.

Besides that, in the current work are a few things, which might need some rework and additions, like:
- take `sameSense` into account by the `addFace`-construction
- the B-Spline surface should be trimmed by the loop boundary: I don't have an idea how to do this task, but I already have a few papers which talk about it.
- discussion about the return type of `convertIfcSurface`: Each surface (e.g., each side of a cube) is calculated into one `Polyhedron`. According to my general understanding of a polyhedron, the entire cube should be stored in one `Polyhedron`-object. Currently, the new function `mergePolyhedronsIntoOnePolyhedron` copies the data into one object.

Core/src/IfcGeometryConverter/FaceConverter.h
pjanck: Good analysis. No need for a special parameter for a `throw`. Your solution seems fine.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -774,20 +774,124 @@ namespace OpenInfraPlatform {
...
+						}
+					}
+				}
+
+				void convertIfcFaceSurface(
+					const EXPRESSReference<typename IfcEntityTypesT::IfcFaceSurface>& faceSurface,
+					std::vector<std::vector<carve::geom::vector<3>>>& faceBoundLoops,
+					const carve::math::Matrix& pos,
+					std::shared_ptr<carve::input::PolyhedronData>& polygon, //Carve polygon of the converted face
+					std::map<std::string, uint32_t> polygonIndices // Contains polygon indices of vertices (x,y,z converted to string)
```
pjanck:
```suggestion
					std::map<std::string, uint32_t>& polygonIndices // Contains polygon indices of vertices (x,y,z converted to string)
```

Don't copy, pass by reference. (apply to other functions as well)

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -774,20 +774,124 @@ namespace OpenInfraPlatform {
...
+				{
+					if (advancedFace.expired()) {
+						throw oip::ReferenceExpiredException(advancedFace);
+					}
+
+					// get reference to IfcFaceBound's (attribute 1)
+					const std::vector<EXPRESSReference<typename IfcEntityTypesT::IfcFaceBound>>& ifcFaceBounds = advancedFace->Bounds;
+					// if one face bound is not of the allowed type: throw; else: continue and use already converted faceBoundLoops
+					if (!std::all_of(ifcFaceBounds.begin(), ifcFaceBounds.end(), [](const auto& ifcFaceBound) {return
+						ifcFaceBound->Bound.isOfType<typename IfcEntityTypesT::IfcEdgeLoop>() || ifcFaceBound->Bound.isOfType<typename IfcEntityTypesT::IfcVertexLoop>(); }))
```
pjanck: Genial!

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -774,20 +774,124 @@ namespace OpenInfraPlatform {
...
+					{
+						throw oip::InconsistentModellingException(advancedFace, "IfcAdvancedFace has a loop type as face boundary which is not allowed.");
+					}
+
+					// get attribute 3: indicates whether the sense of the surface normal agrees with the sense of the topological normal (face bound)
+					// ToDo: take sameSense into account by the addFace-construction
+					const bool sameSense = advancedFace->SameSense;
+
+					// get attribute 2
+					const EXPRESSReference<typename IfcEntityTypesT::IfcSurface>& surface = advancedFace->FaceSurface;
```
pjanck: Is there no `convertIfcSurface` available? If not, please introduce one.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -774,20 +774,124 @@ namespace OpenInfraPlatform {
...
+					{
+						// get surface geometry into inputDataFaceSurface
+						std::shared_ptr<ItemData> inputDataFaceSurface(new ItemData);
+						convertIfcBSplineSurface(surface.as<typename IfcEntityTypesT::IfcBSplineSurface>(), pos, inputDataFaceSurface);
+
+						// ASSUMPTION: the B-Spline-Surface boundary is already coincident with the loop boundary
+						// ToDo: in general case, the B-Spline-Surface has to be trimmed by the loop boundary (or at least checked to be coincident)
+
+						// append surface-faces to target polygon
+						mergePolyhedronsIntoOnePolyhedron(inputDataFaceSurface, polygon);
```
pjanck: Please make this a member function of `ItemData`.

pjanck (commented):
I'm impressed with the screenshots!

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 2034c1c.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: This section is moved into the new function computeIfcFaceSurface (commit b9d8d5d).
Inside there, the surface is needed as EXPRESSReference, because its surface-type (IfcPlane or everything else) is part of the decision about the computation approach. The big if-else construct was simplified in commit 6c3e76d.
Due to this simplification, the existing function `convertIfcSurface` is called.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 237f2b3.

christophKaiser:
Beside small adjustments (commits 2034c1c, bc916f2, b59dcb3, 100c1e1, ef67c56), the structure of functions which are called from `convertIfcFace` was reordered:
A new function `computeIfcFaceSurface` was introduced in commit b9d8d5d. This function is separated, because it is called by `convertIfcFaceSurface` and `convertIfcAdvancedFace` (see calls in commits b9d8d5d and 25d2800) - in both cases it has to do the same computation.
In commit 86f6db6, the parameter `faceBoundLoops` was removed and replaced by the appropriate function call at the location when it is needed.
Previously, the function `mergePolyhedronsIntoOnePolyhedron` has put together several polyhedrons into one polyhedron, but without a topological merge. The result was like a few LEGO bricks in one bowl without sticked together. Now with commit 88e11fe, the existing mesh points are taken into account. If a point (respectively its coordinates) already exists in the target mesh, the existing point will be reused - in this way, several meshes are connected topological. In the example above, the several LEGO bricks in the bowl are sticked together.
Doxygen-comments were added in commit 97efdce.

Commits 34b0b46 and 1cb442c added two functions check whether the points of the boundary loop exist in the surface mesh. If not all boundary loop points are in the surface, a warning is displayed in the console output. Its purpose is to indicate, whether the calculated geometry could be wrong (but don't has to be) - especially in the case when the surface should be trimmed by the boundary (but it isn't, due to a missing functionality).
The example file basin-advanced-brep.ifc works fine with that. However, this warning is raised in the example cube-advanced-brep.ifc, where it shouldn't. It's caused by a default precision of 1.49e-08 (from GeomSettings->getPrecision, fix value in carve), which is to tight. If the precision of 1.0e-4 from this specifc ifc-file is used (defined in the entity `IfcGeometricRepresentationContext`), this problem will not occur - an appropriate ToDo is already in the code:
https://github.com/tumcms/Open-Infra-Platform/blob/025cf1713a7d8a39d24ac58822d7b8c7e7a1aa03/Core/src/IfcGeometryConverter/GeometrySettings.cpp#L102

Beside this aesthetical issue, this pull request has reached its intended functionality (as communicated in dev-meetings). As indicated in the previous section, the trimming of the face surface by the boundary loop is not implemented. This problem will be explained in more detail in a new issue.
Furthermore, the attribute sameSense (see in function computeIfcFaceSurface) is not considered, an appropriate ToDo is left in the code. This attribute affects the normal direction of the faces. At least with the current shading settings in OIP, the normal direction has no visible impact.

Before merge, three ToDo's left on my side:
- [x] Merge `tumcms/development` (with update to VS2019) into this branch, and make sure it works with VS2019 [at last done in commit a83a47b]
- [x] Add unit tests [done in commits 41e84a4, 14e2007 and bbceed0]
- [x] Update the list in SupportedIFCrepresentations.md [done in commit df133e2]

UPDATE 21.02.2022: Updated check boxes of ToDo's.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -726,85 +726,89 @@ namespace OpenInfraPlatform {
...
-				/*! \brief  Converts \c IfcFace to a vector of face vertices.
-				\param	face \c IfcFace entity to be interpreted.
-				\param	pos
-				\param	polygon
-				\param	polygonIndices
+				/*! \brief  Converts \c IfcFace or sub-entity to a vector of face vertices, respectively the mesh of a polyhedron.
+				\param[in]		face				\c IfcFace entity to be converted.
+				\param[in]		pos					The relative location of the origin of the representation's coordinate system within the geometric context.
+				\param[in,out]	polyhedron			Carve polyhedron of the converted face.
+				\param[in,out]	polyhedronIndices	Contains polyhedron indices of vertices (x,y,z converted to string).
```
pjanck: Suggestion: I think Doxy allows for tags like `<see ... >` where you can link to a specific other function in the documentation. Perhaps worth a try here to link to the function that converts the indices?

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -726,85 +726,89 @@ namespace OpenInfraPlatform {
...
+							addArbitraryFaceToPolyhedronData(face, faceBoundLoops, polyhedron, polyhedronIndices);
 						}
 					}
 				}
 
+				/*! \brief  Converts \c IfcFaceSurface or sub-entity to the mesh of a polyhedron.
+				\param[in]		faceSurface			\c IfcFaceSurface entity to be converted.
+				\param[in]		pos					The relative location of the origin of the representation's coordinate system within the geometric context.
+				\param[in,out]	polyhedron			Carve polyhedron of the converted face.
+				\param[in,out]	polyhedronIndices	Contains polyhedron indices of vertices (x,y,z converted to string).
```
pjanck: simile.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -813,20 +817,25 @@ namespace OpenInfraPlatform {
 
 					if (faceSurface.isOfType<typename IfcEntityTypesT::IfcAdvancedFace>())
 					{
-						convertIfcAdvancedFace(faceSurface.as<typename IfcEntityTypesT::IfcAdvancedFace>(), faceBoundLoops, pos, polygon, polygonIndices);
+						convertIfcAdvancedFace(faceSurface.as<typename IfcEntityTypesT::IfcAdvancedFace>(), pos, polyhedron, polyhedronIndices);
```
pjanck: Doesn't the compiler complain about the missing `template` keywords at function calls?

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -835,65 +844,131 @@ namespace OpenInfraPlatform {
...
+					computeIfcFaceSurface(advancedFace, pos, polyhedron, polyhedronIndices);
+				}
+
+				/*! \brief Computes the surface and boundary loop of an \c IfcFaceSurface (or sub-type), and puts them into one polyhedron.
+				\param[in]		ifcFaceSurface		\c IfcFaceSurface entity to be converted.
+				\param[in]		pos					The relative location of the origin of the representation's coordinate system within the geometric context.
+				\param[in,out]	polyhedron			Carve polyhedron of the converted face.
+				\param[in,out]	polyhedronIndices	Contains polyhedron indices of vertices (x,y,z converted to string).
+				*/
+				void computeIfcFaceSurface(
```
pjanck: We have usually used `convert...` for function names that convert Ifc content to carve. For the sake of consistency, please change.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -835,65 +844,131 @@ namespace OpenInfraPlatform {
...
+						std::shared_ptr<ItemData> inputDataFaceSurface = std::make_shared<ItemData>();
+						convertIfcSurface(ifcSurface, pos, inputDataFaceSurface);
 
-						// ASSUMPTION: the B-Spline-Surface boundary is already coincident with the loop boundary
-						// ToDo: in general case, the B-Spline-Surface has to be trimmed by the loop boundary (or at least checked to be coincident)
+						// ASSUMPTION: the Surface boundary is already coincident with the loop boundary
+						// ToDo: in general case, the Surface has to be trimmed by the loop boundary (or at least checked to be coincident)
+						if (!checkBoundaryIsPartOfSurface(inputDataFaceSurface, faceBoundLoops))
+						{
+							BLUE_LOG(warning) << "Surface of IfcFaceSurface (or IfcAdvancedSurface) does not meet the boundary, or the default precision is to tight. Geometry could be wrong!";
```
pjanck:
```suggestion
							BLUE_LOG(warning) << "Surface of IfcFaceSurface (or IfcAdvancedSurface) does not meet the boundary, or the default precision is too tight. Geometry could be wrong!";
```

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -835,65 +844,131 @@ namespace OpenInfraPlatform {
...
+					const std::shared_ptr<ItemData>& itemDataSurface,
+					const std::vector<std::vector<carve::geom::vector<3>>>& faceBoundLoops
+				) const noexcept(true)
+				{
+					// for each point (via std::all_of), check if it is part of the surface;
+					// returns true, if all are part, otherwise false
+					return std::all_of(faceBoundLoops.begin(), faceBoundLoops.end(), [&, this](const auto& faceBoundLoop) {
+						return std::all_of(faceBoundLoop.begin(), faceBoundLoop.end(), [&, this](const auto& pointLoop) {
+							return checkPointIsPartOfSurface(itemDataSurface, pointLoop); });
+					});
```
pjanck: Awesome!!!

pjanck (changes requested):
Some suggestions and praise. Waiting for the rest of the Todos to merge.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: There is already a function `convertIfcFaceSurface` within this pull request, about in line 807.
The idea behind: `convertIfcFaceSurface` will be called from `convertIfcFace` or from somewhere else. Depending on the ifc-entity, that function calls `convertIfcAdvancedFace` or hands over straight to this function `computeIfcFaceSurface`. As well, `convertIfcAdvancedFace` will hand over to `computeIfcFaceSurface`. `computeIfcFaceSurface` performes the actual calculation for both cases, IfcFaceSurface and IfcAdvancedFace. In the visualisation of doxygen, it looks the following:
computeFaceSurface_callStructure
If we would apply the distinction of public and private functions in the geometry converter classes, `convertIfcFaceSurface` and `convertIfcAdvancedFace` would be public, and `computeIfcFaceSurface` would be private.

pjanck: Oh, I see. Please, do use the public/private distinction. You can put the function at the end of the file as a private function - this is already used in some places.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 9cd94a5

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit 0e39d35. I moved it into the protected section, in case another function will inherit from `FaceConverterT`.
Further functions moved to protected with commits 332747a and d6dbb6d.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: On my system, the compiler has not serious issue with this. I'm albe to run the UI-Project and load the example file with an IfcAdvancedFace, as well the Unit Tests run without problems.
However, I still have VS2017 installed on my system - I hope this doesn't affect the behavior of the compiler.

pjanck (commented):
Good work so far!

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: At the moment, there is no explicit function which converts the `polyhedronIndices`. However, thanks for your comment - it showed me the need of refactoring at this point. For another programmer, it isn't easy to see from where the variable `polyhedronIndices` comes from, how it has to be initialised, does it need some kine of default pre-set, and so on.

**A few words about how it works right now**
The variable `std::map<std::string, uint32_t> polyhedronIndices` is declared at one place - in this case inside the function `convertIfcFaceList`. From there it's passed by reference to all functions which work with it.

'Work' refers to the adding a vertex to `polyhedron`. In that case, a `stringstream` with the vertex coordinates is formed and searched inside of `polyhedronIndices`. If the `stringstream` exists, the stored vertex index is used. Otherwise, the vertex is added to `polyhedron` and the `stringstream` is added to `polyhedronIndices`. This task is done by the following sample code:

```cpp
// set string id and search for existing vertex in polyhedron
std::stringstream vertexString;
vertexString << vertex3D.x << " " << vertex3D.y << " " << vertex3D.z;
auto itFound = polyhedronIndices.find(vertexString.str());

uint32_t index = 0;
if (itFound != polyhedronIndices.end()) {
	index = itFound->second;
}
else {
	index = polyhedron->addVertex(vertex3D);
	polyhedronIndices[vertexString.str()] = index;
}

// use index (depends on context)
```

The variable polyhedronIndices is passed by reference through the multiple function calls because its content is filled incrementally and reused in the next calls. In this way, it's responsible for the topological merge of coincident vertices.

The above sample code can be found at multiple places throughout the project - sometimes with little adaptions:
- Subfunctions of `FaceConverterT::convertIfcFaceList`:
    - `FaceConverterT::addTriangleToPolyhedronData`
    - `FaceConverterT::triangulateFace`
    - `ItemData::mergePolyhedronsIntoOnePolyhedron`
- Local use in `FaceConverterT::convertIfcBSplineSurface`
- Local use in `ProfilConverterT::SearchExistingVertex`
- Local use in `SolidModelConverterT::convertIfcSpecificSolidModel` (function is commented out)

**Thoughts about how to refactor**
Because the same task is done at multiple places in the project, it should be reduced into one central function. Within this change, the variable `polyhedronIndices` should be somehow bound to the corresponding polyhedron, hence another programmer don't has to care that much about this variable.

As one conclusion, the information in `polyhedronIndices` is just another representation of the content from the member variable `std::vector<carve::geom3d::Vector> points` of the class `carve::input::PolyhedronData`, respectively `carve::input::PolylineSetData` (`points` is inherited from the class `carve::input::VertexData` where it's declared). Probably, the type `std::map<std::string, uint32_t>` was choosen to have access the `find` function of `std::map`.

To be accurate, `polyhedronIndices` should be a member variable in `VertexData` and the above samle code becomes a memberfunction inside there. However, this is outside of our actual namespace.

Inside the namespace OIP, the class `ItemData` would be the next alternative for the member variable and its depending functions - the function `mergePolyhedronsIntoOnePolyhedron` which works within this topic of topological merging is already there. The first idea was to keep and use the existing instances of `PolyhedronData` and `PolylineSetData`. One disadvantage of this was, that we would need one `polyhedronIndices`-variable for each instance of `PolyhedronData` and `PolylineSetData`, thus four vectors of them (for `closed_polyhedrons`, `open_polyhedrons`, `open_or_closed_polyhedrons` and `polylines`). Futhermore, assuming on function is in execution on one instance of a polyhedron, it will be quite difficult to determine the corresponding `polyhedronIndices`-variable.

A second thought on resolving the determination issue would be the introduction of a new polyhedron instance / variable with its corresponding `polyhedronIndices`-variable. The job of this new instance would be to collect the geometry of one polyhedron; once the polyhedron is complet, its pointer would be handed over into the existing vector `closed_polyhedrons`. The sample code from above would be in the same class, where it can be explicitly called when this functionality should be used; otherwise this variables and functions about topological merging don't have to be called (like in the majority of converter functions which don't use this search methodology).
For the implementation of the new variables and functions could be two options: At first, an extension of the existing class `ItemData` would be an option. The disadvantage on this would be that the new variables and functions would be always visible in an `ItemData`-object.
The second option would be a new class which is used similar to the `ItemData`-class during the collection of the polyhedron. If the collection is finished, the polyhedron would be handed over into the target `ItemData`-object. About the parameter of the converter functions, one issue in this option could be the compatibility between `ItemData`-objects and objects of the new class.

christophKaiser: With reference to todays meeting, the described problem will be subject of a new github-issue.
However, I want to update the doxygen comment to make more clear where the parameter `polyhedronIndices` comes from and how to handle it.

Core/src/IfcGeometryConverter/RepresentationConverter.h
```diff
@@ -513,9 +513,9 @@ namespace OpenInfraPlatform {
 
 				/*! \brief Converts \c IfcTopologicalRepresentationItem to meshes and polylines.
 				 *
-				 * \param[in] mapped_item The \c IfcTopologicalRepresentationItem to be converted.
-				 * \param[in] pos The relative location of the origin of the representation's coordinate system within the geometric context.
-				 * \param[out] itemData A pointer to be filled with the relevant data.
+				 * \param[in] topo_item			The \c IfcTopologicalRepresentationItem to be converted.
+				 * \param[in] objectPlacement	The relative location of the origin of the representation's coordinate system within the geometric context.
+				 * \param[out] itemData			A pointer to be filled with the relevant data.
```
jschlenger, line 518: I know "pos" is maybe not the best name to understand what is meant but I'm quite sure that pos is used in a lot of converter functions 
So having multiple names for the same thing might not be the best idea 
But it's just a minor comment and I leave it up to you if you want to change it

jschlenger (approved):
Looks very good to me.
We should merge this pull request soon.

Core/src/IfcGeometryConverter/RepresentationConverter.h
pjanck: My vote goes for: let's leave at `pos`.

pjanck:
Please mark it as ready for review, then I'll go through the code once more.

christophKaiser:
Commit ac64eb1 updates the Doxygen comments, which are related to the parameters `polyhedron` and `polyhedronIndices`.

A detailed explanation about these parameters is in the description of the function `convertIfcFace`. As shown in the screenshot below, this section is separated with a bold header:
doxygen_description

Furthermore, all functions which have `polyhedron` and `polyhedronIndices` as parameter refer to this description by the Doxygen command `\see`. This generates the section 'See also', as shown in the screenshot:
doxygen_seeAlso

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done.
- New github-issue with the described problem is created.
- Doxygen comments are updated by commit ac64eb1; the result is shown in the comment [below](https://github.com/tumcms/Open-Infra-Platform/pull/462#issuecomment-1046816351).

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done.
Doxygen comments are updated by commit ac64eb1; the result is shown in the comment [below](https://github.com/tumcms/Open-Infra-Platform/pull/462#issuecomment-1046816351).

Core/src/IfcGeometryConverter/RepresentationConverter.h
christophKaiser: I see the point of the more understandable name `objectPlacement`.
With a rough search, VS found 241 matches in 6 files (.Core project). In case we want to change this name, a new PR would be suitable in my opinion.
Anyway, I have made a note to not forget this in the next meeting where we could talk more about this?

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -1685,24 +1690,26 @@ namespace OpenInfraPlatform
...
 				*/
-				void convertIfcEdgeLoop(const EXPRESSReference<typename IfcEntityTypesT::IfcEdgeLoop>& edgeLoop,
-					std::vector<carve::geom::vector<3>>& loop) const noexcept(false)
+				std::vector<carve::geom::vector<3>> convertIfcEdgeLoop(const EXPRESSReference<typename IfcEntityTypesT::IfcEdgeLoop>& edgeLoop
+				) const noexcept(false)
 				{
-					for (auto &orientedEdge : edgeLoop->EdgeList) 
-					{
+					// ToDo: maybe, variable definition of 'loop' has to be moved to correct place
+					std::vector<carve::geom::vector<3>> loop;
```
pjanck: This todo seems resolved.

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -1748,6 +1755,9 @@ namespace OpenInfraPlatform
 							}
 						}
 					}
+
+					// ToDo: maybe, return has to be moved to correct place
+					return loop;
```
pjanck: This todo seems resolved.

Core/src/IfcGeometryConverter/FaceConverter.h
```diff
@@ -726,384 +726,279 @@ namespace OpenInfraPlatform {
...
+				*/
+				bool checkPointIsPartOfSurface(
+					const std::shared_ptr<ItemData>& itemDataSurface,
+					const carve::geom::vector<3>& pointLoop
+				) const noexcept(true)
+				{
+					const std::vector<std::vector<std::shared_ptr<carve::input::PolyhedronData>>>& surfacePolyhedronsCollection{
+						itemDataSurface->closed_polyhedrons,
+						itemDataSurface->open_polyhedrons,
+						itemDataSurface->open_or_closed_polyhedrons };
```
pjanck: I fear that this calls copy-constructor every time the function is called.

Consider making this  `std::vector<std::vector<std::shared_ptr<carve::input::PolyhedronData>>*>`, as in: an array of pointers to ...

Core/src/IfcGeometryConverter/RepresentationConverter.h
pjanck: We can definitely change this name, but rather in a new PR.

Documentation/markdown/SupportedIFCrepresentations.md
```diff
@@ -72,8 +72,8 @@ Meaning of columns:
 | `IfcExtrudedAreaSolidTapered`         | :x:                  | -                          | -                     | -     |
 | `IfcFace`                             | :heavy_check_mark:   | `FaceConverter`            | :heavy_check_mark:    | -     |
 | `IfcFaceBasedSurfaceModel`            | :heavy_check_mark:   | `FaceConverter`            | :heavy_check_mark:    | -     |
-| `IfcFaceBound`                        | :heavy_check_mark:   | `FaceConverter`            | :x:                   | -     |
-| `IfcFaceSurface`                      | :x:                  | -                          | -                     | -     |
+| `IfcFaceBound`                        | :heavy_check_mark:   | `FaceConverter`            | :heavy_check_mark:    | -     |
+| `IfcFaceSurface`                      | :heavy_check_mark:   | -                          | -                     | -     |
```
pjanck:
```suggestion
| `IfcFaceSurface`                      | :heavy_check_mark:   | `FaceConverter`            | -                     | -     |
```

pjanck (approved):
Great work!

Small suggestions, but this can definitely be merged.

jschlenger (approved):
I believe we can merge after the last few comments of pjanck are resolved

Core/src/IfcGeometryConverter/CurveConverter.h
christophKaiser: Originaly by change from return-parameter to return value, I wanted to indicate that the declaration and return was just (roughly) put at the begin and end of the function - I haven't looked carefully into details. However, I will rework this function in the refactoring of issue #528 in a few weeks.
For now, the ToDo is removed with commit 314fe78.

Core/src/IfcGeometryConverter/CurveConverter.h
christophKaiser: Originaly by change from return-parameter to return value, I wanted to indicate that the declaration and return was just (roughly) put at the begin and end of the function - I haven't looked carefully into details. However, I will rework this function in the refactoring of issue #528 in a few weeks.
For now, the ToDo is removed with commit 314fe78.

Core/src/IfcGeometryConverter/FaceConverter.h
christophKaiser: Done in commit c646860.

Documentation/markdown/SupportedIFCrepresentations.md
christophKaiser: Done in commit f621469.

Core/src/IfcGeometryConverter/RepresentationConverter.h
christophKaiser: An issue as reminder and basis for a new PR is created.

pjanck (approved):
Awesome work!

PR #429, opened 2021-04-15
Tree viewer

jschlenger:
Implementation of a viewer that shows the hierarchy tree of IFC files.

PR #465, opened 2021-07-19
Implementing "Zoom in" function for a separate object

### LINKED ISSUE #435: [REQUEST] Implement function “Zoom in” to a separate object in OIP Viewer

Right now, OIP can open several Ifc Files at the same time, thus it is necessary to implement several functionalities, which would help to navigate between those objects. Zoom in to separate object is one of them. 

Create function “Zoom in”, which will zoom in to the object selected.

SamuilsRulovs:
Fixes #435


pjanck (changes requested):
Generally a good start. A few remarks:

- My idea was: the button should be available for each model in the model list on the right (otherwise you cannot possibly know, which object to zoom to). How did you think the selection of the object to zoom to would work (perhaps you have a better idea)?
- why is the icon included twice?
- how to add an icon, please see #387

jschlenger (commented):
Take a look at the first two commits from pull request #429 to get an idea of how to implement what pjanck just suggested.

SamuilsRulovs:
I tried to implement zoom to object function, however something is not working. After clicking on the "Zoom to object" button objects are not moving at all. Maybe my implementation should be a bit more complex. 
image

CMakeLists.txt
```diff
@@ -326,7 +326,7 @@ Option(DOXYGEN_GENERATE_DOCUMENTATION "Build documentation for TUM OpenInfraPlat
 if(DOXYGEN_GENERATE_DOCUMENTATION)
 
 	find_file(DOXYGEN_EXECUTABLE NAMES doxygen.exe HINTS
-		"C:/thirdparty/doxygen"
+		"C:/thirdparty/doxygen/bin"
```
pjanck: Why this change?

pjanck (changes requested):
We need to make this a bit different. Discussing live.

CMakeLists.txt
SamuilsRulovs: I reverted it here: 5c04f9e . I will add an Issue and a new Pull request regarding doxygen on Saturday

SamuilsRulovs:
Now the button works, but function "Zoom to object" is not ready yet (where are several inaccuracies, when we are having more than one object)

SamuilsRulovs:
I have found out, that center of the bounding box is wrong, then we have more than one object.

SamuilsRulovs:
I will have to address this problem first, thus "zoom in" function would work properly

pjanck (commented):
I'm surprised that it still isn't working ...

UserInterface/MainWindow.cpp
```diff
@@ -910,6 +924,29 @@ void OpenInfraPlatform::UserInterface::MainWindow::actionGetCameraState() {
...
+
+	buw::Vector3f zoomMinExtend = modelBBox.min().cast<float>() - delta;
+	buw::Vector3f zoomMaxExtend = modelBBox.max().cast<float>() - delta;
+
+	//buw::Vector3d offset = -zoomBox.center();
+	//buw::Vector3f zoomMinExtend = (zoomBox.min() - fullBBox.center()).cast<float>();
+	//buw::Vector3f zoomMaxExtend = (zoomBox.max() - fullBBox.center()).cast<float>();
+	
+	view_->on_zoomToOneObject(zoomMinExtend, zoomMaxExtend);
+}
```
jschlenger: I feel like the error needs to be somewhere in this function. However it's kind of hard to figure out what's wrong without being able to debug myself and seeing the actual values of the different boxes.
Have you figured out what the right values are for zoomMinExtend and zoomMaxExtent  for a specific example?
I would try to hard code values for theses two until I get the right values so that the zoom is done properly and then reverse engineer how one can get to theses values starting from the given parameters.
Maybe we need another live debugging session to find the error.

SamuilsRulovs: Thank you for the hint. I will try to hardcode these values, thus right now after several trials it is not clear for me either, whether I am trying to calculate right values

SamuilsRulovs:
I went through this Pull Request and concluded that it is impossible to implement the "zoom in" feature into the solution for now. I have created a QPushbutton and a code allowing users to zoom in on objects. Right now, the code works for one object, however, in the case of several objects, the "zoom in" function calculates incorrectly the center of the BBOX. I tried to find a solution previously as well as now, but I couldn't solve it. Therefore I am commenting out my code. (Hopefully, this code can be helpful to others, who will want to continue developing this feature).

jschlenger:
Thank you for taking another look at the pull request. 
I'll close this pull request.

PR #471, opened 2021-09-06
Implementing "delete" function for a separate object

### LINKED ISSUE #436: [REQUEST] Implement function “delete” a separate object in OIP Viewer

Right now, OIP can open several Ifc Files at the same time, thus it is necessary to implement several functionalities, which would help to navigate between those objects. Delete separate object is one of them. 

Create function “Delete”, which will zoom in to the object selected.

SamuilsRulovs:
Fixes #436


UserInterface/MainWindow.cpp
```diff
@@ -910,6 +919,16 @@ void OpenInfraPlatform::UserInterface::MainWindow::actionGetCameraState() {
 	*/
 }
 
+void OpenInfraPlatform::UserInterface::MainWindow::on_actionDeleteObject_triggered(const std::shared_ptr<oip::IModel>& model, const QString& filename){
+	
+	QMessageBox::information(this, tr("Delete object"),
+		tr("The Button works!"), QMessageBox::Ok);
+	//OpenInfraPlatform::Core::DataManagement::Data::removeModel(model);
+
```
jschlenger: I'm quite sure removeModel() will be necessary.
In case it doesn't work you can also try removeallModels(), see if it works and then try to progress from there.

jschlenger: auto& data = OpenInfraPlatform::Core::DataManagement::DocumentManager::getInstance().getData();
	data.removeAllModels();

jschlenger: fileNames -> Datei löschen?

SamuilsRulovs:
I went through this Pull Request and concluded that it is impossible to implement the "delete" feature into the solution for now. 
There is an issue with calling a function altogether (I was trying to implement suggestions from the PR discussion but until the issue with calling a function won't be solved, this feature cannot be implemented.) Therefore I am commenting out my code. (I am not sure if this code should be saved. Maybe it will make more sense to start with implementation from scratch)

jschlenger:
Thank you for having another look at it.
So then I will close this pull request without merging it.

PR #477, opened 2021-10-13
[WIP] Implementing UnitTests for IFC4X3_ RC4

Elvira2227:
- Some unit tests have been added as an example for the version IFC4X3_ RC4

- All tests are missing a `IFCHasAnEssentialEntity` test, please add

- No screenshots were taken. For example, only the one main screenshot is defined for a clothoid, an error is generated for the rest.


jschlenger:
I think it could make sense to start a completely new pull request and branch that directly starts from the development branch and does not include any commits from other branches that are not yet merged (like VS2019 and IfcClothoid). Only copy the changes that you made to the new branch. I think we would have a lot of trouble trying to merge this current branch since it includes things that are part of two other pull requests.

Issue #467, opened 2021-08-12
[REQUEST] New example files OFF / XYZ / BIN


pjanck:
**Describe the solution you'd like**
- wander through https://github.com/pmp-library/pmp-data/tree/master
- see which files OIP can open
- (use ScreenShotMaker to) produce unit tests

PR #479, opened 2021-10-21
Ignoring commented lines for OFF Files

### LINKED ISSUE #382: [BUG] Ignore comments within Off files
**Describe the bug**
Off files can include comment lines starting with "#".
These lines should be ignored in the whole file.
At the moment comment lines are only ignored at the very beginning of the file.

**To Reproduce**
1. Take a off file with comments somewhere in the middle of the file
2. Try to open it

Elvira2227:
Added new function `getNextUncommentedLine` to read only uncommented lines for OFF files

Fixes #382

### DIFF
--- a/Core/src/OffConverter/OffReader.cpp
+++ b/Core/src/OffConverter/OffReader.cpp
@@ -52,11 +52,7 @@ std::shared_ptr<OffModel> OffReader::readFile(const std::string& filename)
 		while (std::getline(offFile, line) && line != "OFF");
 
 		//read number of elements
-		//getline(offFile, line);
-		if (!OffReader::getNextUncommentedLine(offFile, line))
-		{
-			return false;
-		}
+		line = OffReader::getNextUncommentedLine(offFile, line);
 		std::vector<int>	lineData;
 		std::stringstream	lineStream(line);
 
@@ -85,11 +81,7 @@ std::shared_ptr<OffModel> OffReader::readFile(const std::string& filename)
 		for (int i = 0; i < nrOfFaces; i++)
 		{
 			//which type of face (triangle, quad, ...)
-			if (!OffReader::getNextUncommentedLine(offFile, line))
-			{
-				return false;
-			}
-			//std::getline(offFile, line);
+			line = OffReader::getNextUncommentedLine(offFile, line);
 			std::vector<uint32_t> faceVector;
 			std::stringstream lineStream(line);
 
@@ -138,11 +130,7 @@ std::vector<buw::Vector3f> OffReader::readVertices(const int nrOfVertices,
 	std::vector<buw::Vector3f> allVertices;
 	for (int i = 0; i < nrOfVertices; i++)
 	{
-		//std::getline(offFile, line);
-		if (OffReader::getNextUncommentedLine(offFile, line))
-		{
-			line;
-		}
+		line = OffReader::getNextUncommentedLine(offFile, line);
 		buw::Vector3f position;
 		std::stringstream lineStream(line);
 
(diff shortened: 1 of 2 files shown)


Core/src/OffConverter/OffReader.cpp
```diff
@@ -52,7 +52,11 @@ std::shared_ptr<OffModel> OffReader::readFile(const std::string& filename)
 		while (std::getline(offFile, line) && line != "OFF");
 
 		//read number of elements
-		getline(offFile, line);
+		//getline(offFile, line);
+		if (!OffReader::getNextUncommentedLine(offFile, line))
+		{
+			return false;
+		}
```
jschlenger: If I'm not mistaken return false will exit the whole OffReader::readfile function.
I'm quite sure that's not what you try to do. Maybe you need a while loop here?

Core/src/OffConverter/OffReader.cpp
```diff
@@ -81,7 +85,11 @@ std::shared_ptr<OffModel> OffReader::readFile(const std::string& filename)
 		for (int i = 0; i < nrOfFaces; i++)
 		{
 			//which type of face (triangle, quad, ...)
-			std::getline(offFile, line);
+			if (!OffReader::getNextUncommentedLine(offFile, line))
+			{
+				return false;
+			}
```
jschlenger: Same as above

Core/src/OffConverter/OffReader.cpp
```diff
@@ -130,7 +138,11 @@ std::vector<buw::Vector3f> OffReader::readVertices(const int nrOfVertices,
 	std::vector<buw::Vector3f> allVertices;
 	for (int i = 0; i < nrOfVertices; i++)
 	{
-		std::getline(offFile, line);
+		//std::getline(offFile, line);
+		if (OffReader::getNextUncommentedLine(offFile, line))
+		{
+			line;
+		}
```
jschlenger: line; doesn't do anything

Core/src/OffConverter/OffReader.cpp
```diff
@@ -241,5 +253,20 @@ buw::Vector3f OffReader::calcNormal(const buw::Vector3f& vertex1,
...
+bool OffReader::getNextUncommentedLine(std::ifstream& offFile, std::string& line)
+{
+	while (std::getline(offFile, line))
+	{
+		for (int i = 0; i < line.size(); i++)
+		{
+			if (line[i] != '#')
+			{
+				return true;
+			}
```
jschlenger: The first time a character is != # this function will stop completely and return true.
You also have to be sure to handle lines like "1.0  0.0 1.4142 #comment" right.
Even though this contains a comment we need all the information up until the # sign

jschlenger (commented):
I suggest you use the example from here https://en.wikipedia.org/wiki/OFF_(file_format) , delete the color values for now but still leave the comments #red, #green, ...
Store it in a file and try to open it. Step through the code with the debugger to check where your code doesn't behave as expected.

Core/src/OffConverter/OffReader.cpp
Elvira2227: Yes, you are right. In the last commit, I decided to read the code this way (but not quite right), but then I saw that our code in any case reads only a certain number of first characters (for example, for the first 4 characters from the line) and therefore will not a priori read comments at the end of the line.
Another story is when a comment is left accidentally or deliberately inside the points like "1 #comment 1 0". Perhaps we should eliminate this option too?

Core/src/OffConverter/OffReader.cpp
Elvira2227: To check it, I took our file from the testdata and added randomly comments. I checked the code with debugs and it really doesn't count comments. 
This new function already iterates in its body and returns only the line that is free of comments. 
For example:
image

Core/src/OffConverter/OffReader.cpp
jschlenger: The way theses comments are usually handled is that starting from the # sign the rest of the line is treated as a comment because otherwise it would be necessary to have a second sign to be able to tell where the comment ends 
I guess that off does not allow having comments in the middle of the line as you described it

jschlenger:
Regarding the color question in the live meeting. I just took a look at the OffReader code. Using the same color value for all of the vertices should definitely be fine since the Off format defines one color per face.

jschlenger:
Once you got everything working you could add a unit test for a off file with comments and colors. You can also test the off. files mentioned in issue #467

jschlenger (commented):
Good work!

PR #484, opened 2021-10-29
Implement color reader for OFF Files

### LINKED ISSUE #354: [REQUEST] Reading colors from .OFF files
The .off file importing doesn't support reading colors yet. Add this functionality.

Adapt OffReader.cpp accordingly 

**Additional context**
There are multiple ways in which colors can be defined in the .off file format.
Be aware of all of them before implementing a solution that only supports one of them!

Elvira2227:
Fixes #354 

Added new function `readColorsFromFace` to read colors from OFF Files. 

In the attachment there are some screenshots to represent colors read from the cube.off 
I am not sure if that representation is correct. I am confused by the blur at the edges. Maybe this is a feature of our UI?

image
image


Elvira2227:
The code still supports 3 colors: red, green and blue. I would suggest adding a few more exceptions to the code to avoid reading errors.

Core/src/OffConverter/OffReader.cpp
```diff
@@ -98,6 +98,11 @@ std::shared_ptr<OffModel> OffReader::readFile(const std::string& filename)
 			{
 				readQuadFace(lineStream, indices, buwVertices, offVertices);
 			}
+			//read color
+			else if (faceType == 5)
+			{
+				readColorsFromFace(lineStream);
+			}
```
jschlenger: facetype is meant to distinguish between triangles, quadrilaterals, and faces with more edges 
this is always defined by the first value in the lines that define the faces 
At the moment only triangles and quadrilaterals are supported, so the code will never enter your else if 
So reading color doesn't have anything to do with facetype but it could make sense to replace it with a variable that tells if the file/line has color values to be read or not

Core/src/OffConverter/OffReader.cpp
```diff
@@ -238,7 +251,23 @@ buw::Vector3f OffReader::calcNormal(const buw::Vector3f& vertex1,
...
+	lineStream >> colorVector[5] >> colorVector[6] >> colorVector[7];
+	//RGB Red #FF0000 255 000 000 -> 1.0f, 0.0f, 0.0f
+	//RGB Green #00FF00 000 255 000 -> 0.0f, 1.0f, 0.0f
+	//RGB Blue #0000FF 000 000 255-> 0.0f, 0.0f, 1.0f
+	if (colorVector[5] == 255)//red
+		colorVector.x = 1.0f;
+	else if (colorVector[6] ==255)//green
+		colorVector.y = 1.0f;
+	else if (colorVector[7] == 255)//blue
+		colorVector.z = 1.0f;
```
jschlenger: There are many more possibilities than just red, green and blue. Basically all three color values can range from 0.0f to 1.0f making it possible to show any possible color. So the values in the off file can range from 0 to 255 and all of them should be considered. It should be possible to add a simple function that converts the values from 0 to 255 to values from 0.0f to 1.0f accordingly

Core/src/OffConverter/OffReader.cpp
```diff
@@ -202,20 +210,25 @@ void OffReader::readQuadFace(std::stringstream& lineStream,
...
-	buw::Vector3f color1(0.0f, 0.0f, 1.0f); //default color; to be changed later on 
-	buw::Vector3f color2(0.0f, 0.0f, 1.0f); //default color; to be changed later on 
+	//read colors
+	//buw::Vector3f color1(0.0f, 0.0f, 1.0f); //default color; to be changed later on 
+	//buw::Vector3f color2(0.0f, 0.0f, 1.0f); //default color; to be changed later on 
+	//buw::Vector3f color3(0.0f, 0.0f, 1.0f); //default color; to be changed later on
+
+	buw::Vector3f color1 = OffReader::readColorsFromFace(lineStream);
+	buw::Vector3f color2 = OffReader::readColorsFromFace(lineStream);
+	buw::Vector3f color3 = OffReader::readColorsFromFace(lineStream);
```
jschlenger: I think it would make sense to move this into your newly created readColorsFromFace function

Core/src/OffConverter/OffReader.cpp
```diff
@@ -163,12 +168,15 @@ void OffReader::readTriangleFace(std::stringstream& lineStream,
 	buw::Vector3f normal = calcNormal(vector1, vector2, vector3);
 
 	//read color
-	buw::Vector3f color(0.0f, 0.0f, 1.0f); //default color; to be changed later on 
+	buw::Vector3f color1 = OffReader::readColorsFromFace(lineStream);
+	buw::Vector3f color2 = OffReader::readColorsFromFace(lineStream);
+	buw::Vector3f color3 = OffReader::readColorsFromFace(lineStream);
```
jschlenger: Move this into the readColorsFromFace function

Core/src/OffConverter/OffReader.cpp
```diff
@@ -163,12 +168,15 @@ void OffReader::readTriangleFace(std::stringstream& lineStream,
...
+	//buw::Vector3f color(0.0f, 0.0f, 1.0f); //default color; to be changed later on 
 
 	//create vertices with position, color and normal and add to list of all vertices
-	vertices.push_back(buw::VertexPosition3Color3Normal3(vector1, color, normal));
-	vertices.push_back(buw::VertexPosition3Color3Normal3(vector2, color, normal));
-	vertices.push_back(buw::VertexPosition3Color3Normal3(vector3, color, normal));
+	vertices.push_back(buw::VertexPosition3Color3Normal3(vector1, color1, normal));
+	vertices.push_back(buw::VertexPosition3Color3Normal3(vector2, color2, normal));
+	vertices.push_back(buw::VertexPosition3Color3Normal3(vector3, color3, normal));
```
jschlenger: I think there are some mistakes here which could be also the reason why the colors of your cube in the screenshot are quite black and faded. 
So I guess the way this should be is that VertexPosition3Color3Normal3 takes a  color vector that includes all the color values specified in the lines so e.g. 0.0f 1.0f 0.0f all together. All three vectors need the same color vector so that the complete face has one uniform color.

Core/src/OffConverter/OffReader.cpp
```diff
@@ -202,20 +210,25 @@ void OffReader::readQuadFace(std::stringstream& lineStream,
...
+	buw::Vector3f color2 = OffReader::readColorsFromFace(lineStream);
+	buw::Vector3f color3 = OffReader::readColorsFromFace(lineStream);
 
 	//create vertices with position, color and normal and add to list of all vertices
 	//first triangle
 	vertices.push_back(buw::VertexPosition3Color3Normal3(vector1, color1, normal1));
-	vertices.push_back(buw::VertexPosition3Color3Normal3(vector2, color1, normal1));
-	vertices.push_back(buw::VertexPosition3Color3Normal3(vector3, color1, normal1));
+	vertices.push_back(buw::VertexPosition3Color3Normal3(vector2, color2, normal1));
+	vertices.push_back(buw::VertexPosition3Color3Normal3(vector3, color3, normal1));
```
jschlenger: Same as above

Core/src/OffConverter/OffReader.cpp
```diff
@@ -202,20 +210,25 @@ void OffReader::readQuadFace(std::stringstream& lineStream,
...
 
 	//second triangle
 	vertices.push_back(buw::VertexPosition3Color3Normal3(vector3, color2, normal2));
-	vertices.push_back(buw::VertexPosition3Color3Normal3(vector4, color2, normal2));
-	vertices.push_back(buw::VertexPosition3Color3Normal3(vector1, color2, normal2));
+	vertices.push_back(buw::VertexPosition3Color3Normal3(vector4, color3, normal2));
+	vertices.push_back(buw::VertexPosition3Color3Normal3(vector1, color1, normal2));
```
jschlenger: Same as above

testdata/Off/cube.off
```diff
@@ -12,9 +12,9 @@ OFF
...
-4 4 6 7 5
-4 0 4 5 1
-4 1 5 7 3
-4 0 2 6 4
+4 0 1 3 2 255 0 0 #red
+4 2 3 7 6 0 255 0 #green
+4 4 6 7 5 0 0 255 #blue
+4 0 4 5 1 0 255 0
+4 1 5 7 3 0 0 255
+4 0 2 6 4 255 0 0
```
jschlenger: Later on also try with random values between 0 and 255

jschlenger (commented):
If something is not clear we can also have a Zoom meeting next week

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved f419556

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved f419556

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved f419556
I just found a mistake in a function and now the colors are displayed fine
image

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved f419556

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved f419556

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved fb61590
I tried to just delete this part of the code and everything works fine

testdata/Off/cube.off
Elvira2227: Resolved fb61590

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved fb61590
I tried to just comment this part of the code and everything works fine. I tried also to check other colors for example yellow and it looks perfect :)
image

jschlenger (commented):
Good progress. I suggest cleaning up the code by removing the comments that were just part of the older version of the code that was not working properly. Also try to finish the pull request regarding ignoring comments and then we can merge both requests at the same time. Unless you have something more in mind to do here.

jschlenger (commented):
Have you also checked that OFF files that do not provide any colors can be read without any issues?
I will take a look at the issue with only a limited amount of colors that can be displayed and see if that's something we can easily fix. If it looks like a bigger task I'll merge this pull request and create a new issue for the color problem.

jschlenger:
I haven't found anything that would prevent colors to be represented accordingly. 
I could imagine that the issue is that the color vector is expecting values between 0.0f and 1.0f and we are providing it with values between 0 and 255 which could lead to the result that everything that is not 0 is interpreted as 1.0f. Maybe take a look at that and see if color values between 0.0 and 1.0 give the results that we need.

Core/src/OffConverter/OffReader.cpp
```diff
@@ -163,12 +164,10 @@ void OffReader::readTriangleFace(std::stringstream& lineStream,
 	buw::Vector3f normal = calcNormal(vector1, vector2, vector3);
 
 	//read color
-	buw::Vector3f color(0.0f, 0.0f, 1.0f); //default color; to be changed later on 
+	buw::Vector3f color = OffReader::readColorsFromFace(lineStream);
 
 	//create vertices with position, color and normal and add to list of all vertices
 	vertices.push_back(buw::VertexPosition3Color3Normal3(vector1, color, normal));
-	vertices.push_back(buw::VertexPosition3Color3Normal3(vector2, color, normal));
-	vertices.push_back(buw::VertexPosition3Color3Normal3(vector3, color, normal));
```
pjanck: Each vertex needs a color and a normal - this is probably the reason to see darker edges on the faces.

Core/src/OffConverter/OffReader.cpp
```diff
@@ -238,7 +236,14 @@ buw::Vector3f OffReader::calcNormal(const buw::Vector3f& vertex1,
...
-	return buw::Vector3f(normal.x, normal.y, normal.z);
+	return buw::Vector3f(normal.x, normal.y, normal.z);	
+}
+buw::Vector3f OffReader::readColorsFromFace(std::stringstream& lineStream)
+{
+	carve::geom::vector<3> colorVector;
+	lineStream >> colorVector[0] >> colorVector[1] >> colorVector[2];
+
+	return buw::Vector3f(colorVector.x, colorVector.y, colorVector.z);
 }
```
pjanck: How does this behave if no color is provided in the line?

Core/src/OffConverter/OffReader.cpp
```diff
@@ -238,7 +236,14 @@ buw::Vector3f OffReader::calcNormal(const buw::Vector3f& vertex1,
...
 
-	return buw::Vector3f(normal.x, normal.y, normal.z);
+	return buw::Vector3f(normal.x, normal.y, normal.z);	
+}
+buw::Vector3f OffReader::readColorsFromFace(std::stringstream& lineStream)
+{
+	carve::geom::vector<3> colorVector;
+	lineStream >> colorVector[0] >> colorVector[1] >> colorVector[2];
+
+	return buw::Vector3f(colorVector.x, colorVector.y, colorVector.z);
```
pjanck: To normalize this, divide each number with 255.

pjanck (changes requested):
Suggestions.

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved bbb7fb7

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved bbb7fb7

Core/src/OffConverter/OffReader.cpp
Elvira2227: I deleted it by mistake
Resolved in bbb7fb7

Elvira2227:
At the last commit bbb7fb7 I added all suggestions and it finally works with color representation (i mean all of rgb color, see some screenshots)
image
image

jschlenger (commented):
Looks good. I believe we can merge this pull request now.

Core/src/OffConverter/OffReader.cpp
```diff
@@ -98,6 +98,7 @@ std::shared_ptr<OffModel> OffReader::readFile(const std::string& filename)
 			{
 				readQuadFace(lineStream, indices, buwVertices, offVertices);
 			}
+			
 			else
 			{
```
pjanck: No need for the additional line between the end of `if` and start of the corresponding `else`.

Core/src/OffConverter/OffReader.cpp
```diff
@@ -238,7 +238,18 @@ buw::Vector3f OffReader::calcNormal(const buw::Vector3f& vertex1,
...
+buw::Vector3f OffReader::readColorsFromFace(std::stringstream& lineStream)
+{
+	carve::geom::vector<3> colorVector;
+	lineStream >> colorVector[0] >> colorVector[1] >> colorVector[2];
+
+	if (!colorVector.exactlyZero())
+		return buw::Vector3f(colorVector.x/255.0f, colorVector.y/255.0f, colorVector.z/255.0f);
+	else
+		return buw::Vector3f(0.0f, 0.0f, 1.0f); // default color - blue
+
```
pjanck: I imagine this doesn't work, if the color should be black. Since then the values are all 0, thus making the `colorVector` being `exactlyZero` - and falling in the `else` making the color blue. Have you tried coloring an object black?

This needs a better solution. Proposal: Check if there is anything left in the `lineStream` - if not, then go for default, otherwise calculate the color.

testdata/Off/cube.off
```diff
@@ -12,9 +12,9 @@ OFF
...
-4 4 6 7 5
-4 0 4 5 1
-4 1 5 7 3
-4 0 2 6 4
+4 0 1 3 2 128 0 255 # lila
+4 2 3 7 6 255 255 0 # yellow
+4 4 6 7 5 0 0 255 # blue
+4 0 4 5 1 255 191 0 # orange
+4 1 5 7 3 128 255 0 # ligth green
+4 0 2 6 4 255 0 64 # pink
```
pjanck: Love it!

pjanck (changes requested):
I'm not entirely sure this would work as intended for all cases. Please take a look - perhaps no change in code is required.

Core/src/OffConverter/OffReader.cpp
jschlenger: TUM Blue!
default

Elvira2227:
In the last commit 368de21 was added correctly implementation of black color and defaul TUM-blue color. See screen shots
image

UPDATE: In b9f0a2f corrected conditions of if statement. It works!

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved 368de21

Core/src/OffConverter/OffReader.cpp
Elvira2227: Resolved 368de21

jschlenger (approved):
Looks good!

PR #485, opened 2021-11-05
Unit tests for OFF Files

### LINKED ISSUE #381: [REQUEST] New Unit Tests for the Off file format
So far there is only one Off unit test existing.
The testdata folder includes a folder with more Off example files.
The task is to write a unit test for every example file.

**Additional context**
The helix unit test can be used as a guidance to create similar tests for the other files.

Elvira2227:
Fixes #381 

New unit tests for the .off files are added

NOTE: Producing screen shots is in progress


pjanck (changes requested):
The `cube` example does not have any screen shots. Is that a mistake or intended?

UnitTests/Off/boxcube/src/boxcube.cpp
```diff
@@ -0,0 +1,186 @@
...
+};
+
+TEST_F(Boxcube, AllEdgesAndFacesRed)
+{
+	// check for the number of vertices 
+	// does not coincide with the number of vertices mentioned in the off file since some vertices are
+	// stored once for every triangle they are part of
+	EXPECT_THAT(model->geometry().vertices.size(), Eq(324));
+	// check for the number of indices (see comment for vertices)
+	EXPECT_THAT(model->geometry().indices.size(), Eq(324));
```
jschlenger, line 63: Are you sure about the numbers here?
The file defines 48 vertices so I wonder how we end up with 324

UnitTests/Off/boxtorus/src/boxtorus.cpp
```diff
@@ -0,0 +1,186 @@
...
+};
+
+TEST_F(Boxtorus, AllEdgesAndFacesRed)
+{
+	// check for the number of vertices 
+	// does not coincide with the number of vertices mentioned in the off file since some vertices are
+	// stored once for every triangle they are part of
+	EXPECT_THAT(model->geometry().vertices.size(), Eq(192));
+	// check for the number of indices (see comment for vertices)
+	EXPECT_THAT(model->geometry().indices.size(), Eq(192));
```
jschlenger, line 63: same

UnitTests/Off/cone/src/cone.cpp
```diff
@@ -0,0 +1,186 @@
...
+};
+
+TEST_F(Cone, AllEdgesAndFacesRed)
+{
+	// check for the number of vertices 
+	// does not coincide with the number of vertices mentioned in the off file since some vertices are
+	// stored once for every triangle they are part of
+	EXPECT_THAT(model->geometry().vertices.size(), Eq(120));
+	// check for the number of indices (see comment for vertices)
+	EXPECT_THAT(model->geometry().indices.size(), Eq(120));
```
jschlenger, line 63: same (and for the rest of the UTs as well)

jschlenger (commented):
All of the numbers of expected vertices seemed odd to me but maybe there is a valid reason for that.

Elvira2227:
Regarding cube. I am recieving an Exception in Line 303 in D3D11RenderSystem.cpp 
image

We with @SamuilsRulovs think, that it is something to do with sub variable, because it is empty. 
image

pjanck:
And not with the others? Weird. This shouldn't happen, not in the BlueFramework backend.

I see several possibilities:

- is the source code of the unit test's .cpp different then that of other passing unit tests?
- is the cube example file somehow polluted?

UnitTests/Off/boxcube/src/boxcube.cpp
Elvira2227: I made it because when debugging it failed the Test. And actually the value of this Test producing in console (see scr)
image

Elvira2227:
In the  last commit f2f70db added new  unit test for colored cube. Producing scr's for new one only after merging #484

@pjanck I sudgest that the problem is in the low file size. (The  code and the .off file is usual and hav'nt any difference  then that other. I  have produced  new unit  test for  colored cube  and it compiled without any  problem. 
image

NOTE: After merging #484 we need update all screen shots ,because the default color was  changed in "TUM-blue"

Elvira2227:
In the last commit d42fc1c were  colors for all unit tests updated. Cube have still problem with producing screen shots. Should it deleted from unit tests (we have also same geometry "cube_colored" without any problems)?

jschlenger:
Yes, I suggest to delete it.

jschlenger (approved):
Looks good. 
We are able to merge now.

PR #521, opened 2022-01-13
Fix placement converter

jschlenger:
The getPlane() function of the placement converter was not working anymore because of minor changes of IfcPlacement in IFC4x3_RC4.
The complete function has been refactored.

### DIFF
--- a/Core/src/IfcGeometryConverter/PlacementConverter.h
+++ b/Core/src/IfcGeometryConverter/PlacementConverter.h
@@ -1448,57 +1448,21 @@ namespace OpenInfraPlatform {
 
                 // Function 6: Get Plane. 
 				// \internal TODO refactoring
-				static void getPlane(const std::shared_ptr<typename IfcEntityTypesT::IfcAxis2Placement3D> axis2placement3d,
+                void getPlane(const EXPRESSReference<typename IfcEntityTypesT::IfcAxis2Placement3D> axis2placement3d,
                     carve::geom::plane<3>& plane,
-                    carve::geom::vector<3>& translate,
-                    double length_factor)
+                    carve::geom::vector<3>& translate)
                 {
-                        carve::geom::vector<3>  location(carve::geom::VECTOR(0.0, 0.0, 0.0));
-                        carve::geom::vector<3>  local_x(carve::geom::VECTOR(1.0, 0.0, 0.0));
-                        carve::geom::vector<3>  local_y(carve::geom::VECTOR(0.0, 1.0, 0.0));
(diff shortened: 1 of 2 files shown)


Core/src/IfcGeometryConverter/SolidModelConverter.h
```diff
@@ -2235,7 +2235,7 @@ namespace OpenInfraPlatform
 				carve::geom::vector<3> baseSurfacePosition;
 				carve::math::Matrix basePositionMatrix(carve::math::Matrix::IDENT());
 				
-				placementConverter->getPlane(elemBaseSurface->Position.lock(), baseSurfacePlane, baseSurfacePosition, this->UnitConvert()->getLengthInMeterFactor());
+				placementConverter->getPlane(elemBaseSurface->Position.lock(), baseSurfacePlane, baseSurfacePosition);
```
pjanck: You can remove the `.lock()` since we're dealing with `EXPRESSReference` now. Apply appropriately for the whole PR.

Core/src/IfcGeometryConverter/PlacementConverter.h
```diff
@@ -1448,57 +1448,18 @@ namespace OpenInfraPlatform {
...
-                            }
-                        }
-
-                        local_z.normalize();
-
-                        carve::geom::plane<3> p(local_z, location);
-                        plane.d = p.d;
-                        plane.N = local_z;
-                        translate = location;
+                    carve::math::Matrix placement = convertIfcPlacement(axis2placement3d);
```
pjanck: If you know, that it is an Axis2Placement3D, then call the function that takes it - optimization.

Core/src/IfcGeometryConverter/SolidModelConverter.h
pjanck: Although, I'm unsure if that is the syntax for optionals. Let's have a call if it is not obvious.

pjanck (changes requested):
Awesome simplifications!

Core/src/IfcGeometryConverter/PlacementConverter.h
jschlenger: Resolved: 
978ac11

Core/src/IfcGeometryConverter/SolidModelConverter.h
jschlenger: Everything is working fine without "lock()"

pjanck (approved):

PR #522, opened 2022-01-27
Correcting a minor mistake

SamuilsRulovs:


jschlenger (approved):

pjanck:
Can you also rather use atan2 instead of atan?

jschlenger (approved):

pjanck (approved):

PR #523, opened 2022-02-06
Incorporating LAS Reader into solution

### LINKED ISSUE #502: [BUG] LAS file format is not supported, OIP shows nothing
**Describe the bug**
LAS is not supported by OIP, thus nothing is shown in OIP viewer.

**To Reproduce**

1. In CMAKE select OIP_WITH_POINT_CLOUD_PROCESSING and generate solution anew
2. Build OpenInfraPlatform.PointCloudProcessing
3. Start OIP
4. Open .las file
5. See error

**Expected behavior**
OIP should support .las files

**Screen shot**
image

SamuilsRulovs:
Fixes #502


jschlenger (commented):
Did this already allow you to successfully use the liblas library?
At some point the documentation should be adapted so that everyone knows that he or she has to run these cmd files.

pjanck (commented):
I'm not convinced by the solution. We need to discuss this. Especially since CloudCompare code is getting linked in our solution on code basis and we compile it in our solution. So I don't understand the need to compile it with a *cmd*.

jschlenger:
Should we have a meeting next week on this? (This week my schedule is already quite packed)

SamuilsRulovs:
Yes, it would be great, thank you!

SamuilsRulovs:
I have included LAS into our solution and uncommented code, which stands for processing LAS. However, it causes the error with boost to pop up 
image

PointCloudProcessing/CMakeLists.txt
```diff
@@ -120,6 +126,7 @@ set(DEPENDENT_INCLUDE_DIRS
 	#${GeoTIFF_INCLUDE_DIR}
 	#${dep_libgeo_SOURCE_DIR}/include
 	#${dep_liblas_SOURCE_DIR}/include
+	${LAS_INCLUDE_DIR}/include
```
jschlenger: Are you sure about this one?
Shouldn't it be {libLAS_INCLUDE_DIR}?
After the cmake configuration process you should be able to see all defined variables 
Check if LAS_INCLUDE_DIR existis

PointCloudProcessing/CMakeLists.txt
```diff
@@ -140,6 +147,7 @@ target_link_libraries(OpenInfraPlatform.PointCloudProcessing PUBLIC
 	    #${TIFF_LIBRARIES}
 	    #${GeoTIFF_LIBRARIES}
 		#geotiff_library
+		${LAS_LIBRARIES}
```
jschlenger: Same as above

jschlenger (commented):
Check the things that I mentioned in the code. Also check if there are any warnings regarding liblas. Does cmake say a specific version of liblas has been found?
Also try to put the code that is using liblas in comments and see if you still get the error.
This or similar things could help to narrow down the problem.

PointCloudProcessing/CMakeLists.txt
SamuilsRulovs: Resolved: [f25f5ee](https://github.com/tumcms/Open-Infra-Platform/pull/523/commits/f25f5ee654502282c595af54562fcfc50caaffbd)

PointCloudProcessing/CMakeLists.txt
SamuilsRulovs: Resolved: [f25f5ee](https://github.com/tumcms/Open-Infra-Platform/pull/523/commits/f25f5ee654502282c595af54562fcfc50caaffbd)

SamuilsRulovs:
image
After correcting a variable to libLAS_INCLUDE_DIR I received a lot of Errors from CMake during generation process

Every error is saying, that something from boost is not found

However boost is found and if I switch to development everything works
image

pjanck:
> Every error is saying, that something from boost is not found

This could be a problem in boost version that is required by liblas and the oip. Have you checked the versions?

SamuilsRulovs:
If I checked correctly, liblas doesn't use boost at all

SamuilsRulovs:
I couldn’t solve the issue of libLAS causing OIP generation to crash. For this reason, I am commenting out all code created for libLAS incorporation. I will also create an Issue describing everything in detail.

jschlenger (commented):
Since we didn't mange to get this to work properly lets close this pull request.

PR #527, opened 2022-02-11
Error Message updated

Elvira2227:
This pull request resolved a problem with the detection error while reading an IFC File. Now it shows the number of row which contain an error.


jschlenger (approved):
Good change!

pjanck (approved):
Lftm

PR #536, opened 2022-04-08
Adjustment of Doxygen documentation (enable)

### LINKED ISSUE #535: [BUG] GenerateDocumentation (Doxygen) is not included in the VS-project
**Describe the bug**
After _Configure_ and _Generate_ by CMake with default settings from `CMakeLists.txt`, the project `OpenInfraPlatform.GenerateDocumentation` isn't included / visible in the folder _Commands_ of Visual Studio.

**To Reproduce**
1. In CMake, delete cache
2. Click _Configure_, _Generate_ and _Open Project_ in CMake
3. Search in the Solution Explorer of Visual Studio (folder _OpenInfraPlatform/Commands_, project `OpenInfraPlatform.GenerateDocumentation` -> it won't be there)

**Expected behavior**
The documentation "Doxygen for Open Infra Platform" states, that Doxygen will be enabled by default:
https://github.com/tumcms/Open-Infra-Platform/blob/4fa2e0a4d39c42ae113ff4ba90e04e9b380d1ed3/Documentation/markdown/DoxygenHelp.md?plain=1#L24

However, Doxygen is disabled in the top-level CMakeLists.txt:
https://github.com/tumcms/Open-Infra-Platform/blob/4fa2e0a4d39c42ae113ff4ba90e04e9b380d1ed3/CMakeLists.txt#L324

Of course, the user can switch on this option in CMake. But I don't know, if an unexperienced user would find this option.
At least one file should be adjusted to be consistent.

**Suggestion**
I found the commit, which has turned off the setting in the CMakesLists.txt. Unfortunately, to me it isn't obvious why it was turned off.

On my system, I tested the CMake build process today:
1. Changed the above mentioned CMakeLists.txt to
   `Option(DOXYGEN_GENERATE_DOCUMENTATION "Build documentation for TUM OpenInfraPlatform." ON)`
2. Deleted cache in CMake
3. Clicked _Configure_, _Generate_ and _Open Project_ in CMake
4. Built `OpenInfraPlatform.GenerateDocumentation` in Visual Studio
5. The generated documentation opend automatically in a browser, like expected without noticeable side effects

If there is a background issue which is unknown to me, I would withdraw my suggestion.

christophKaiser:
Fixes #535, detailed description with test of the suggested change in the issue.

### DIFF
--- a/Documentation/markdown/DoxygenHelp.md
+++ b/Documentation/markdown/DoxygenHelp.md
@@ -30,16 +30,16 @@ If `DOXYGEN_GENERATE_DOCUMENTATION` is checked in the CMake GUI (default = OFF),
 Additionally, the following options can be selected:
 - `DOXYGEN_INCLUDE_COMMENTED_ONLY` (default = ON): If selected, only the documentation for explicitly commented entities will be generated. Otherwise, doxygen will assume that all entities are documented, even if there is no documentation available. 
 - `DOXYGEN_INCLUDE_INTERNAL` (default = OFF): If selected, this will include all comments preceded by "\internal". These are comments intended for developers only. So if the generated documentation is intended for external users, this option should not be selected.
-- DOXYGEN_INCLUDE_EARLYBINDING (default = OFF): If selected (... and already built), this will include the selected IFC schemas in the documentation. Otherwise, the schemas will not be included in the documentation, but this does not influence the selection of IFC schemas (see EARLYBINDING_ tags in CMake GUI).
-- DOXYGEN_AUTO_OPEN_DOCUMENTATION (default = ON): If selected, the landing page of the documentation will automatically open. If not selected, you will have to find and open the generated documentation manually (see section below on building the documentation). 
+- `DOXYGEN_INCLUDE_EARLYBINDING` (default = OFF): If selected (... and already built), this will include the selected IFC schemas in the documentation. Otherwise, the schemas will not be included in the documentation, but this does not influence the selection of IFC schemas (see EARLYBINDING_ tags in CMake GUI).
+- `DOXYGEN_AUTO_OPEN_DOCUMENTATION` (default = ON): If selected, the landing page of the documentation will automatically open. If not selected, you will have to find and open the generated documentation manually (see section below on building the documentation). 
 
 ### ConfigureDoxygen.cmake
-If DOXYGEN_GENERATE_DOCUMENTATION is selected in CMake, CMake will include ConfigureDoxygen.cmake. This file includes all settings for the generation of doxygen documentation for the Open Infra Platform project. Changes to documentation settings should be made here. doxygen_add_docs() creates the GenerateDocumentation target. The up-to-date version of the file can be found here:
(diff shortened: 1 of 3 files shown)


pjanck (changes requested):
While I agree with the problem, I'd fix the other way around - fix the documentation, with a clear instruction, where to turn it on. 

Reasoning: building documentation takes quite some time and hinders "build solution" greatly.

christophKaiser:
Commit aae0c68 is only an optional suggestion from my side. At the end of the setup process, probably not everyone takes a detailed look into the Doyxgen guidelines.

Documentation/markdown/DoxygenHelp.md
```diff
@@ -19,23 +19,27 @@ Refer to the [doxygen manual](http://www.doxygen.nl/manual) for everything else.
 
 ## Doxygen settings in CMake
 
+By default, doxygen is disabled. 
+To enable doxygen, search for the `DOXYGEN_GENERATE_DOCUMENTATION` option in CMake and set the check box.
```
pjanck, line 23: Since we're revisiting this in this PR, I'd propose to prepend the variable with `OIP_` so that it is grouped together with other OIP's variables. Please change accordingly everywhere, where it comes up (I assume there are mentions in the Doxygen documentation as well).

Documentation/markdown/DoxygenHelp.md
```diff
@@ -19,23 +19,27 @@ Refer to the [doxygen manual](http://www.doxygen.nl/manual) for everything else.
...
-- DOXYGEN_INCLUDE_EARLYBINDING (default = OFF): If selected (... and already built), this will include the selected IFC schemas in the documentation. Otherwise, the schemas will not be included in the documentation, but this does not influence the selection of IFC schemas (see EARLYBINDING_ tags in CMake GUI).
-- DOXYGEN_AUTO_OPEN_DOCUMENTATION (default = ON): If selected, the landing page of the documentation will automatically open. If not selected, you will have to find and open the generated documentation manually (see section below on building the documentation). 
+- `DOXYGEN_INCLUDE_EARLYBINDING` (default = OFF): If selected (... and already built), this will include the selected IFC schemas in the documentation. Otherwise, the schemas will not be included in the documentation, but this does not influence the selection of IFC schemas (see EARLYBINDING_ tags in CMake GUI).
+- `DOXYGEN_AUTO_OPEN_DOCUMENTATION` (default = ON): If selected, the landing page of the documentation will automatically open. If not selected, you will have to find and open the generated documentation manually (see section below on building the documentation).
```
pjanck, line 34: Apply here as well (prepend with `OIP_`).

Documentation/markdown/SetupHelp.md
```diff
@@ -177,6 +177,7 @@ Build the project **OpenInfraPlatform.GenerateDocumentation** within *Commands*
 *NOTE:* This step may take quite some time.
 
 Read more about Doxygen in our [guidelines](./DoxygenHelp.md).
+*NOTE:* Doxygen is disabled by default. See the guidelines for a description how to enable Doxygen.
```
pjanck, line 180:
```suggestion
*NOTE:* Documentation generation is disabled by default. See the guidelines for a description of how to enable Doxygen.
```

pjanck (changes requested):

pjanck:
Since we're renaming already, I'd call these variables `OIP_DOCUMENTATION_...`, not mentioning doxygen (it may change in the future).

jschlenger (commented):
Nothing to add from my side.

Issue #524, opened 2022-02-08
[REQUEST]Implementing corresponding functions for IfcSpline


Elvira2227:
**Is your feature request related to a problem? Please describe.**
In #478 new unit tests were implemented for version IFC4X3_RC4 . For "bloss", "helmert", "sine", "cosine", "cubic", "viennese bend" curves we need to add corresponing functions to support them.

**Describe the solution you'd like**
The function convertIfcCurveSegment shoul be updated and new overloads for getPointOnCurve and getDirectionOfCurve should be added for:

- [ ] Sine curve (IfcSine) #553 
- [ ] Cosine curve (IfcCosine)

The function for reading `IfcThirdOrderPolynomialSpiral` should be added for:
- [ ] Bloss curve #550 

The function for reading `IfcSeventhOrderPolynomialSpiral` should be added for:
- [ ] Viennese Bend curve #552 

The function for reading `IfcSecondOrderPolynomialSpiral` should be added for:
- [ ] Helmert curve #551 

The function for reading `IfcPolyonmialCurve` should be added for: 
- [ ] Cubic curve #539 

**Additional context**
For examples of implementation please see #494 and #482

pjanck:
Great issue description!

A note to the developer: for *first-*, *second-*, and so on *-orderpolynomial*, a common function can be called by all with zeros at the corresponding locations.

PR #539, opened 2022-04-16
IfcPolynomialCurve

Elvira2227:
In this pull request implementation was added for `IfcPolynomialCurve`. 
The following functions were added in `CurveConverter.h`:
1. `getPointOnCurve`
2. `getDirectionOfCurve` - here we need to implement additional code in `covertIfcCurveSegment` to calculate rotation around x,y axis and also rotation in 3D. Now we can only rotate around axis z. 
3. `calculatePolynomialCurve` - to calculate polynomial from coefficietnts

For explanation there is a screen shot of cubic curve with coefficientsX (0., 1., 0.) and coefficientsY (0., 0., 0.01, 0.).

image

Note: 
This is related to #524 
Please see starting from commit [2b83b5b](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/2b83b5b8c7cf4fd4ab989610f5bd12c8ff54b114)


UnitTests/Schemas/IFC4X3_RC4/bloss-curve_100.0_300_1000_1_Meter/src/bloss-curve_100.0_300_1000_1_Meter.cpp
```diff
@@ -0,0 +1,100 @@
...
+TEST_F(BlossCurve, ImageIsSaved)
+{
+	// Arrange
+	buw::Image4b image = rendererIfc->captureImage();
+
+	// Act
+	buw::storeImage(testPath("bloss-curve_100.0_300_1000_1_Meter.png").string(), image);
+
+	// Assert
+	EXPECT_NO_THROW(buw::loadImage4b(testPath("bloss-curve_100.0_300_1000_1_Meter.png").string()));
```
pjanck, line 83: I can't seem to see this file commited.

UnitTests/Schemas/IFC4X3_RC4/clothoid_100.0_1000_-300_1_Meter/src/clothoid_100.0_1000_-300_1_Meter.cpp
```diff
@@ -0,0 +1,98 @@
...
+	EXPECT_NE(result2, express_model->entities.end());
+}
+
+TEST_F(Clothoid, ImageIsSaved)
+{
+	// Arrange
+	buw::Image4b image = rendererIfc->captureImage();
+
+	// Act
+	buw::storeImage(testPath("clothoid_100.0_1000_-300_1_Meter.png").string(), image);
```
pjanck, line 78: The file with this name does not need be committed (only the top view is needed for these unit tests). Apply for all unit tests.

pjanck (commented):
General remarks:

- some unit tests are still missing a screen shot
- it seems weird to have so many unit tests for `line` that seem all equal
- some `png` are not needed as they are not used in checking

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,6 +2901,189 @@ namespace OpenInfraPlatform
...
+					
+					// calculate angle between two polynomial curves
+					//double angle = std::atan2(angleY, angleX);
+
+					if (polynomialConstantCntX>0 && polynomialConstantCntY>0)
+					{
+						double angle = std::atan2(angleY, angleX);
+						return carve::geom::VECTOR(std::cos(angle), std::sin(angle), 0.);
+
+					}
```
jschlenger: Are you sure that you need to insert angleX and angleY here?
From a quick view it would feel more natural to insert pointX and pointY into atan to get an angle than using two angles to get another angle.
Please take a closer look here to see if what you've done produces the results that you expect.

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,6 +2901,189 @@ namespace OpenInfraPlatform
...
+					// Only works if correctly rotation around y axis is in convertIfcCurveSegment 
+					else if (polynomialConstantCntX>0 && polynomialConstantCntZ>0)
+					{
+						double angle = std::atan2(angleZ, angleX);
+						return carve::geom::VECTOR(std::cos(angle), 0., -std::sin(angle));
+					}
```
jschlenger: Same as above

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,6 +2901,189 @@ namespace OpenInfraPlatform
...
+					// Only works if correctly rotation around x axis is in convertIfcCurveSegment 
+					else if (polynomialConstantCntY>0 && polynomialConstantCntZ>0)
+					{
+						double angle = std::atan2(angleZ, angleY);
+						return carve::geom::VECTOR(0., std::cos(angle), std::sin(angle));
+					}
```
jschlenger: Same as above

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,6 +2901,189 @@ namespace OpenInfraPlatform
...
+				{
+					return getPointOnCurve(polynomialCurve, parameter * this->UnitConvert()->getLengthInMeterFactor());
+				}
+				
+				carve::geom::vector<3> getPointOnCurve(const EXPRESSReference<typename IfcEntityTypesT::IfcPolynomialCurve>& polynomialCurve,
+					const double& parameter) const noexcept(false)
+
+				{
+
+					std::vector<typename IfcEntityTypesT::IfcReal> coefficientX, coefficientY, coefficientZ;
```
jschlenger: it would make more sense to use plural here to make it clear that these variables include multiple coefficients

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,6 +2901,189 @@ namespace OpenInfraPlatform
...
+					typename IfcEntityTypesT::IfcReal sum = coefficient[0];
+					double factor = 1;
+
+					for (int i = 1; i < n; i++)
+					{
+						factor *= length;
+						sum += coefficient[i] * factor;
+					}
+					return sum;
+				}
```
jschlenger: If you'd use peters functions here I think you would need XbyAngleDeviationPolynomialByTerms (and the same for y) and not XbyAngleDeviationPolynomial, because the first calls the second so we would miss a step by using the second one.

jschlenger (commented):
I'm really not sure if your angle calculation is right.
Please double check and test with break points if the produced values make sense.

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: I think you are right. 
I thought that we can find separatelly the angles for X and Y polynomial. Thats why I used the same function from SpiralUtils as for the clothoids. 
My suggestion to use the implementation from this sourse : https://firebirdsql.org/refdocs/langrefupd21-intfunc-atan2.html
Then we can use `atan2(sin(angleY), cos(angleX))`. 
See cf59b36

Elvira2227:
There are several changes in the last commits cf59b36 and 28f6421.

- First, I was trying to find corresponding functions from the SpiralUtils, but it returned high values of X and Y. Thats why I desided to write a separate functions which calculates integral by using numerical methods. That will helps to find appropriate values for X and Y.
- New functions `integrateParameter` , `calculatePolynomialDerivative`, `intagral` have been placed in `CurveConverter.h`.
- I have noticed that `integrateParameter` returned the value with the approximation error 0.4%. 
- In the commit 28f6421 are small corrections for `integrateParameter` function.
- I tested my code with polynomialConstantX = (0.,1.,0.) and polynomialConstantY = (0.,0.,1., 0.)

Here is the implemented polycurve:
image

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -3026,36 +3027,59 @@ namespace OpenInfraPlatform
...
 						coefficientX = polynomialCurve->CoefficientsX;
 						polynomialConstantCntX = std::size(coefficientX);
-						// Implement polynomial term for x coordinate
-						x = calculatePolynomialCurve(coefficientX, parameter);
+						//std::vector<double> polynomialConstantX;
+						//convert to double
+						for (int i = 0; i < std::size(coefficientX); i++)
+						{
+							polynomialConstantX.push_back(double(coefficientX[i]));
+						}
```
pjanck: Learning possibility: look up std::transform function, which could help you achieve the same with less code. (there are some usages in OIP code already)

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -3064,27 +3088,110 @@ namespace OpenInfraPlatform
...
 
-				/*! \brief Calculates a polinomial curve.
+				/*! \brief Calculates a polinomial curve
 				* \param[in] coefficient			A vector with coeffitients corresponding ordinate from \c IfcPolynomialCurve.
 				* \param[in] parameter				A length during the curve from \c IfcCurveSegment.
 				* \return							The location of the ordinate point.
 				* \note
 				*/
-				typename IfcEntityTypesT::IfcReal calculatePolynomialCurve(const std::vector<typename IfcEntityTypesT::IfcReal> coefficient, double length) const noexcept(false)
+				typename IfcEntityTypesT::IfcReal calculatePolynomialCurve(std::vector<double> coefficient, const double parameter) const noexcept(false)
```
pjanck:
```suggestion
				typename IfcEntityTypesT::IfcReal calculatePolynomialCurve(std::vector<double>& coefficient, const double parameter) const noexcept(false)
```

If you're not changing the values of the array, always pass it by reference, not by value (saves time and RAM).

pjanck (commented):
Some suggestions :)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Thank you for your advice! It is really faster than usual.
Resolved: 98adeb1

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved: [98adeb1](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/98adeb1aa6baac4fd3b81995623cb7bff780ddb0)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: In this commit was added a new implementation for resolving this problem. Unfortunately, there are problems with Peter's function, it returns very high values. 
Resolved [cf59b36](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/cf59b361c5e046defaad80ee290c0a19ad2cefbd)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: I have tasted different cases and this is a result that presents the first data of vector `segmentDirections`

angle = atan2(std::sin(angleY),std::cos(angleX));
carve::geom::vector<3> v1 = carve::geom::VECTOR(std::cos(angle), std::sin(angle), 0. );

**[0]:{1.0000000000000000, 0.0000000000000000, 0.0000000000000000}- corresponds to point (0,0,0)**
[1]:{0.93699400295766744, 0.34934544282324143, 0.0000000000000000}
[2]:{-0.67306627244558215, 0.73958217453925257, 0.0000000000000000}
...
angle = atan2(y,x);
carve::geom::vector<3> v2 = carve::geom::VECTOR(std::cos(angle), std::sin(angle), 0. );

**[0]:{1.0000000000000000, 0.0000000000000000, 0.0000000000000000} - corresponds to point (0,0,0)**
[1]:{0.78609999999999047, 0.61795320999998504, 0.0000000000000000}
[2]: {1.2495999999999947, 1.5615001599999867, 0.0000000000000000}
...

It only says that we have similar values for different calculations. Interestingly, our rotation matrix uses only the first point (segmentDirections[0]) to detect the direction. I think this should be the next task to check over. 

P.S: I decided to place your suggestion about angle calculations.

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved: [98adeb1](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/98adeb1aa6baac4fd3b81995623cb7bff780ddb0)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved: [98adeb1](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/98adeb1aa6baac4fd3b81995623cb7bff780ddb0)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved: [98adeb1](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/98adeb1aa6baac4fd3b81995623cb7bff780ddb0)

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -34,6 +34,7 @@
 #include "SplineConverter.h"
 
 #include "BlueFramework/Core/Diagnostics/log.h"
+#include <fstream>
```
pjanck: Why do we need fstream here?

pjanck (commented):
Need to review the code more thoroughly, just one quick comment.

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: I forgot to delete this line after I tested some functionality.  
Resolved: [aac48ed](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/aac48ed9f897985601b33b741f9ace040cb8f854)

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,7 +2901,378 @@ namespace OpenInfraPlatform
...
+					return getDirectionOfCurve(polynomialCurve, parameter * this->UnitConvert()->getLengthInMeterFactor());
+				}
+				template<>
+				carve::geom::vector<3> getDirectionOfCurve(const EXPRESSReference<typename IfcEntityTypesT::IfcPolynomialCurve>& polynomialCurve,
+					const double& parameter) const noexcept(false)
+				{				
+					//std::vector<typename IfcEntityTypesT::IfcReal> coefficientsX, coefficientsY, coefficientsZ;
+					//std::vector<double> polynomialConstantX, polynomialConstantY, polynomialConstantZ;
+					int polynomialConstantCntX, polynomialConstantCntY, polynomialConstantCntZ;
+					std::vector<double> polynomialConstantX, polynomialConstantY, polynomialConstantZ;
```
jschlenger: I would use plural for all of them
polynomialConstantsX, ...

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,7 +2901,378 @@ namespace OpenInfraPlatform
...
+						std::transform(coefficientsZ.begin(), coefficientsZ.end(), polynomialConstantZ.begin(), [](auto it) { return it; });
+						polynomialConstantCntZ = std::size(polynomialConstantZ);
+						//calculate angle
+						//angleZ = SpiralUtils::AngleByAngleDeviationPolynomial(polynomialConstantZ, polynomialConstantCntZ, parameter);
+					}
+					
+					// calculate angle between two polynomial curves
+					//double angle = std::atan2(angleY, angleX);
+
+					if (polynomialConstantCntX>0 && polynomialConstantCntY>0)
```
jschlenger: Is the polynomialConstantCnt only used for these if clauses?
If that's the case one could save a couple of lines by just writing
if (polynomialCurve->CoefficientsX && polynomialCurve->CoefficientsY)
then the polynomialConstCnt could be removed completely
However, this is just a personal preference and can also be left the way it is right now

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,7 +2901,378 @@ namespace OpenInfraPlatform
...
+					// Impplement different cases for 2D and 3D
+					if (polynomialConstantCntX > 0 && polynomialConstantCntY > 0)
+					{
+						double t = integrateParameter(polynomialConstantX, polynomialConstantY, parameter);
+						x = calculatePolynomialCurve(polynomialConstantX, t);
+						y = calculatePolynomialCurve(polynomialConstantY, t);
+						return carve::geom::VECTOR(x, y, 0.);
+					}
+					
+					else if (polynomialConstantCntX > 0 && polynomialConstantCntZ > 0)
```
jschlenger: same as above (constantCnt)

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,7 +2901,378 @@ namespace OpenInfraPlatform
...
+				carve::geom::vector<3> getPointOnCurve(const EXPRESSReference<typename IfcEntityTypesT::IfcPolynomialCurve>& polynomialCurve,
+					const double& parameter) const noexcept(false)
 
+				{
+
+					std::vector<typename IfcEntityTypesT::IfcReal> coefficientsX, coefficientsY, coefficientsZ;
+					//typename IfcEntityTypesT::IfcReal x, y, z;
+					double x, y, z;
+					int polynomialConstantCntX, polynomialConstantCntY, polynomialConstantCntZ;
+					std::vector<double> polynomialConstantX, polynomialConstantY, polynomialConstantZ;
```
jschlenger: plural

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,7 +2901,378 @@ namespace OpenInfraPlatform
...
+						// Interpret coefficients for X
+						coefficientsX = polynomialCurve->CoefficientsX;
+						// convert to double
+						polynomialConstantX.resize(coefficientsX.size());
+						std::transform(coefficientsX.begin(), coefficientsX.end(), polynomialConstantX.begin(), [](auto it) { return it; });
+						polynomialConstantCntX = std::size(polynomialConstantX);
+						//x = SpiralUtils::XbyAngleDeviationPolynomial(polynomialConstantX, polynomialConstantCntX, parameter);
+						//x = SpiralUtils::XbyAngleDeviationPolynomialByTerms( 0., 0., 0., 0., 0., 0., 1., 0., parameter);
+
+					}
```
jschlenger: Could it make sense to put this coefficient x y z retrieval into a separate function that is then called by getPointOnCurve and by getDirectionOfCurve?
Looks like the same code is used at least twice which could be avoided

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,7 +2901,378 @@ namespace OpenInfraPlatform
...
+				}
+
+				/*! \brief Calculates a polinomial curve
+				* \param[in] coefficient			A vector with coeffitients corresponding ordinate from \c IfcPolynomialCurve.
+				* \param[in] parameter				A length during the curve from \c IfcCurveSegment.
+				* \return							The location of the ordinate point.
+				* \note
+				*/
+				typename IfcEntityTypesT::IfcReal calculatePolynomialCurve(std::vector<double>& coefficient, const double& parameter) const noexcept(false)
```
jschlenger:
```suggestion
				typename IfcEntityTypesT::IfcReal calculatePolynomialCurve(std::vector<double>& coefficients, const double& parameter) const noexcept(false)

```

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,7 +2901,378 @@ namespace OpenInfraPlatform
...
+					{
+						factor *= parameter;
+						sum += coefficient[i] * factor;
+					}
+					return sum;
+				}
+
+				/*! \brief Calculates derivative of the polynomial curve
+				* \param[in] polynomialConstant	    A vector with coeffitients corresponding ordinate from \c IfcPolynomialCurve.
+				* \param[in] b				        Parameter value from intagrateParameter
```
jschlenger:
```suggestion
				* \param[in] b				        Parameter value from integrateParameter

```

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,7 +2901,378 @@ namespace OpenInfraPlatform
...
+					int polynomialConstantCnt = std::size(polynomialConstant);
+					double value = polynomialConstant[1];
+					int factor = 2;
+					// derivative of the polynomial
+					for (int i = 2; i < polynomialConstantCnt; i++)
+					{
+						value += polynomialConstant[i] * factor * b;
+						b *= b;
+						factor += 1;
+					}
```
jschlenger: There might be a mistake here but I'm not sure
Lets say we want to have the derivative of the function 2 * b which would be 2 
It seems that this is not taken into account in this for loop because everything is multiplied with b 
Is this "first step" skipped on purpose?

Core/src/IfcGeometryConverter/CurveConverter.h
```diff
@@ -2895,7 +2901,378 @@ namespace OpenInfraPlatform
...
+				* \note
+				*/
+				double integrateParameter(std::vector<double>& polynomialConstantX, std::vector<double>& polynomialConstantY, const double& length) const noexcept(false)
+				{
+					double a = 0.;
+					double b = 0.;
+					int n = 1;// small numbers return better values
+					double value = 0.;
+
+					if (length == 0) { return b; }
```
jschlenger, line 3165: in the following lines the curly braces are not used when returning b 
Try to be consistent and use them everywhere or leave them out completely

jschlenger (commented):
A couple of minor comments here and there

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: I use polynomialConstantCntXYZ to check if the coefficients exist. 
Because in the beginning, I create polynomialConstantXYZ vectors and then check it in this if statement it should always return true.
If `polynomialConstantXYZ` exist and its size `polynomialConstantCntXYZ `return positive value.

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved [102578f](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/102578f6c6f89ea5cb99c511be0ab4b573f31c0a)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved [102578f](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/102578f6c6f89ea5cb99c511be0ab4b573f31c0a)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved [102578f](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/102578f6c6f89ea5cb99c511be0ab4b573f31c0a)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved [102578f](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/102578f6c6f89ea5cb99c511be0ab4b573f31c0a)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved [102578f](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/102578f6c6f89ea5cb99c511be0ab4b573f31c0a)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Thank you for you comment! 
In the code is only one mistake(!!), it does not work with vectors with size 1.
In this case we have constant polynomial with derivative equal 0.

For example:
*Example vector:*
**p3** = (1,2,3,4)
**polynomial** = 1 + 2*b + 3*b*b + 4*b*b*b
**derivative by hand** = 2 + 6*b + 12*b*b

I skiped first step because the first member of the polynomialConstant vector is linear term which derivative is eqaul to polynomialConstant[1]
And thats why I start with the second iteration .

*With code:*
value = 2;
faktor = 2;
`i = 2`
**value** = 2 + 3*2*b 
b = b*b
faktor = 3
`i = 3`
**value** = 2 + 3*2*b + 4*3*b*b
b = b*b*b
faktor = 4
**returns derivative with value** 2 + 3*2*b + 4*3*b*b == 2 + 6*b + 12*b*b
Resolved in [102578f](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/102578f6c6f89ea5cb99c511be0ab4b573f31c0a)

Core/src/IfcGeometryConverter/CurveConverter.h
Elvira2227: Resolved in [963fd76](https://github.com/tumcms/Open-Infra-Platform/pull/539/commits/963fd7698ce4e515c750642c3673d0cb6cf9f35b)

pjanck (commented):
Not there yet, right?

PR #550, opened 2022-06-13
IfcThirdOrderPolynomialSpiral

Elvira2227:
In this pull request I implemented functions for reading the `IfcThirdOrderPolynomialSpiral`.
Now we can implement "Bloss curve" and this pull request can be addressed to issue #524.
In the CurveConverter.h I added:
1. `getPointOnCurve()` 
2. `getDirectionOfCurve()`
3. additional statement in `convertIfcCurveSegment()`

Here you can see the screen shot (it presents BlossCurve_100.0_300_1000_1_Meter) :
image


pjanck (approved):
I assume that unit tests work correctly.

jschlenger (commented):
Have you checked that you are not only visualizing the curve that Peter added to the IFC file to visually check if the calculation is done right?
(as was the case for IfcSine in the beginning)

jschlenger:
Everything seems fine. Let's merge this.

PR #551, opened 2022-06-13
IfcSecondOrderPolynomial

Elvira2227:
In this pull request I implemented functions for reading the `IfcSecondOrderPolynomialSpiral`.
Now we can implement "Helmert curve" and this pull request can be addressed to issue #524.
In the CurveConverter.h I added:
1. `getPointOnCurve()` 
2. `getDirectionOfCurve()`
3. additional statement in `convertIfcCurveSegment()`

Here you can see the screen shot (it presents HelmertCurve_100.0_300_1000_1_Meter) :
image


pjanck (approved):
I assume that unit tests work correctly.

jschlenger (commented):
Have you checked that you are not only visualizing the curve that Peter added to the IFC file to visually check if the calculation is done right?
(as was the case for IfcSine in the beginning)

jschlenger:
It seems like quite some changes that were made for IfcSine are reverted in this pull request, which is most probably not intended. Also some bspline sreenshots that were change recently are reverted here.
I think that should not be the case. Maybe pulling the most recent changes from the development branch could help.
Let me know in case there are any issues, it seems like you tried to merge but then reverted it.

jschlenger (approved):
Lftm

PR #556, opened 2022-07-12
Changing arguments of IFCSIUNIT from $ to *  in linear-placement.ifc

### LINKED ISSUE #76: Example file "linear-placement" is not working.

SamuilsRulovs:
I guess it fixes #76 

After that minor change OIP can show example file, however I do not know, linear-placement.ifc is shown correctly, since there are no photos, how this example file should look like. 
image

Continues #447


pjanck (commented):
That's a start. However, those vertical drops are definitely wrong. Why and how they occur is beyond me.

I also think that it is not necessary to fix this, since the supported schema should be IFC4x3 and the IFC4x1 has been retired; with the example files being updated at some point (https://github.com/buildingSMART/IFC4.3.x-development/issues/491#issuecomment-1162715247).

jschlenger:
Thanks for the hint.
Let's focus on IFC4x3.

jschlenger (commented):
Thanks for the hint.
Let's focus on IFC4x3.

pjanck:
Propose to merge this and open a new issue revolving around the wrong visualization.

pjanck (approved):

Issue #560, opened 2022-07-29
Slow calculation time for IfcSine and IfcCosine


Elvira2227:
**Describe the bug**
Calculation of `IfcSine` and `IfcCosine` takes a long time.

**To Reproduce**
1. Go to `CurveConverter.h`
2. Scroll down to the line 2693 for `IfcSine` or to the line 2748 for `IfcCosine`
3. Set an additional breakpoint if necessary
4. Wait for the end of the calculation

**CMake configuration**
If applicable, list any special configurations you have selected in CMake.
- EARLYBINDING_WITH_IFC4X3_RC4: selected

**Additional context**
On my machine it took approximately 9 minutes.

pjanck:
Well that's just how the calculation is organized. Improvements are definitely possible. But I would argue that this currently has a low priority.

PR #561, opened 2022-07-29
IfcCosine

Elvira2227:
In this pull request adds new implementation for `IfcCosine`.
The sturcture of the functions `getpointOnCurve` and `getDirectionOnCurve` are similar with representation of IfcSpline.

**Note:**
There are a mistake in allready merged functions of `IfcSine`, for instance in the declaration of the template functions `getPointOnCurve` and `getDirectionOfCurve`. Previously, the code did not reached this functions and it represents geometry only with cartesianPointList #565 . Now this mistake has been fixed. 
Another point corresponds to issue #560. The calculation takes quite a long time.

In the example file `cosine-curve_100.0_300_1000_1_Meter` were are not 100 point as usual, but 101. Therefore, our implementation does not reach the last point. I think, it is a bug in example file.
Here is a screenshot of this example. And we see that the maximum point, which should be the last point, is just the penultimate point.
image

ToDo:

- [x] Add contexte to `ReleaseNotes`
- [x] Add context to `SupportedIFCrepresentations`


jschlenger (commented):
Looks good so far.
Did you already work on the issue with the long calculation times?

pjanck (approved):
lftm

Is there an issue/PR for the unit tests?

PR #562, opened 2022-08-01
as there is only Qt 5.15.2 in the installer - add 5.15.2 to compile

mach0:


pjanck (commented):
I do not understand the need for this change. I've downloaded the installer of Qt and can select 5.15.1:

image

Is there a specific reason for this?

jschlenger:
It is important to check the box **Archive** to be able to see Qt versions other than the most recent one.
I also just checked the installer and for me it is still possible to download version 5.15.1.

mach0:
hi,
very strange, I downloaded the Qt installer twice and tried several times to get 5.15.1 (to follow the manual closely) but I could not get 5.15.1. Maybe I overlooked something, or maybe it is platform dependent which versions you will see? I don't know. I just thought (as I only got 5.15.2) it might affect others too. BTW: Is there a specific reason to select MinGW 8.1.0 32bit? Because I could only compile it with installing MSVC and MinGW 32 and 64 bit (being on a 64bit OS).
But after all I am no cpp expert. Just got it compiling by adding those lines.

pjanck:
Interesting. 

> Just got it compiling by adding those lines.

This is great! No need to downgrade to 5.15.1, then. 

> Is there a specific reason to select MinGW 8.1.0 32bit

Not that I'm aware of. If it works with 64bit, I'm happy. This is definitely an area of OIP that was not explored in depth; we just went with the version where we were sure it worked.

pjanck (approved):
LFTM

If possible: Please add a note to the SetupHelp documentation that 5.15.2 works as well.

jschlenger:
The written part of the documentation asks to select the 64bit version.
There is a mistake in the image provided. I'll replace the image very soon.

mach0:
thats quite interesting.
Just downloaded the Qt Installer again (this time within Linux) and .. only 5.15.2 available regardless of what I am selecting (Archive/LTS/...) there is no change here, and no 5.15.1.
Regarding the MinGW 8.1.0 32bit - I tried it here (as to follow the Manual closely) and it did NOT work unless I installed the MinGW 64bit Option .. And there was also a errorline (unfortunately cannot remember exactly what it was) that brought me to that path - it was something with could not find 64bit version .... - but surely something with 64bit).
Now I am stuck with the UI saying something like "invalid settings file" during startup - but as far as I can see there is only one setting.xml - and it is an valid xml .. but I guess that would be something for an forum question

jschlenger:
Are you pressing the **Filter** button after applying changes to the category boxes?
Only once you do that the changes will be applied.

pjanck:
> only 5.15.2 available

I was not aware of that. This spurs the idea of updating the whole Qt stuff to 6.x version and modernizing a thing or two.

> Now I am stuck with the UI saying something like "invalid settings file" during startup.

I'd need to see that; I can't remember having this problem in the past. My email should be in one of the PDF papers in Documentation folder.

mach0:
> 

thanks a lot for that hint - not very obvious I must say (as I asked 2 other persons if they can find the 5.15.1 version and nobody even came close to the idea of pressing filter after selection). But thanks a lot for that hint.
Regarding Qt 6.x I am sure you know the support for 5.15 will end with May 2023 (with 2 year extension for subscribers). If I can help with compiling and testing (and maybe extending documentation) I think Qt6 would be very welcome.

I will try the whole compile again with 5.15.1 to see if it changes something on the "invalid setting" error during startup, but I am afraid it should have nothing to do with Qt Versions.

pjanck:
> I will try the whole compile again with 5.15.1 to see if it changes something on the "invalid setting" error during startup, but I am afraid it should have nothing to do with Qt Versions.

I propose to retire (and not merge) this PR if the error gets resolved. Otherwise I'm fine with merging.

jschlenger:
I'm glad my hint helped.
I'll update the documentation to make it clear that the changes in the category section are only applied once the user presses the **Filter** button.

mach0:
Switching to Qt 5.15.1 did not help, so compiling is successfull with both versions and I guess one could also use 5.15.2.

As the "self compiling" manual stopped at this point, I tried to start OpenInfraPlatform.UI.exe from the Release Folder with "An unexpected error occured: Invalid settings file" " 

Forgive me if I am doing/starting something wrong here, but there is no hint what to try now.

As I don't want to fill up a pull request with subsequent questions, please point me to page with "sustainable" communication - Forum or mailinglist or something like that?

jschlenger:
I always start the OIP from Visual Studio no matter if I'm in release or debug mode.
Actually I tried to run it by executing the exe file which results in the same exact error that you get.
Being completely honest I don't exactly know the reason for that but I guess there might be some dependency related issues that aren't completely resolved when executing the exe file. I can just recommend to start it from Visual Studio which workers perfectly fine for me.

jschlenger:
The easiest might be to contact me via email.
You can find it here: https://www.cms.bgu.tum.de/de/team/jonas-schlenger 
I'm glad to help with any further issues related to the installation process or other problems.

pjanck:
> I tried to start OpenInfraPlatform.UI.exe from the Release Folder with "An unexpected error occured: Invalid settings file" "

Have you compiled the *CopyOpenInfraPlatformUIResources* project?

> Forgive me if I am doing/starting something wrong here, but there is no hint what to try now.

No worries! Been there, done that. You are most welcome to post any questions and concerns. We're constantly improving the documentation, so the more errors we can put answered in our FAQ, the better.

> please point me to page with "sustainable" communication

Shoot me / @jschlenger an email and we can invite you to our Slack.

PR #563, opened 2022-08-02
Update Qt_Installation_settings.png

mach0:
correct wrong 32 bit selection screenshot


pjanck (approved):
Thank you for the updated screen shot!

